PR #21, opened 2023-10-18
Use Github Releases to Check for Version Upgrade

foodprocessor:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Reset cloudfuse version to 0.1.0.
Check github releases for latest version for upgrade message instead of using the blobfuse docker container repo

### Checklist

- [x] Tested locally
- [x] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### DIFF
--- a/common/version.go
+++ b/common/version.go
@@ -31,6 +31,7 @@ import (
 	"strings"
 )
 
+const CloudfuseReleaseURL = "https://api.github.com/repos/Seagate/cloudfuse/releases"
 const CloudfuseListContainerURL = "https://blobfuse2.blob.core.windows.net/release"
 const CloudfuseWarningsURL = "https://aka.ms/blobfuse2warnings"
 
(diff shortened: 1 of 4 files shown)


common/version.go
```diff
@@ -31,6 +31,7 @@ import (
 	"strings"
 )
 
+const CloudfuseReleaseURL = "https://api.github.com/repos/Seagate/cloudfuse/releases"
 const CloudfuseListContainerURL = "https://blobfuse2.blob.core.windows.net/release"
```
jfantinhardesty, line 35: Should we remove these two lines, since they are for blobfuse2. Maybe could be a second PR?

jfantinhardesty (approved):
LGTM

common/version.go
foodprocessor: Yeah, this neighborhood needs cleaning up. I'll see what I can remove without starting too much trouble.

foodprocessor: Ok, just did that.

PR #29, opened 2023-10-24
Loop over pages to get requested object count.

foodprocessor:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

ReadDir on S3 would only return up to 1k entries.
Now List() iterates until it reaches the 5k limit.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/s3storage/s3storage.go
+++ b/component/s3storage/s3storage.go
@@ -238,9 +238,9 @@ func (s3 *S3Storage) StreamDir(options internal.StreamDirOptions) ([]*internal.O
 	path := formatListDirName(options.Name)
 	var iteration int                   // = 0
 	var marker *string = &options.Token // = nil
-	var entriesFetched int32
-	for {
-		newList, newMarker, err := s3.storage.List(path, marker, options.Count-entriesFetched)
+	var totalEntriesFetched int32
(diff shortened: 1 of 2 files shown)


jfantinhardesty (approved):

jfantinhardesty:
I'm wondering if it would be better to change StreamDir in s3storage to loop over the call to list, rather than changing how list works. Since the problem is really that StreamDir doesn't loop over list.

foodprocessor:
> I'm wondering if it would be better to change StreamDir in s3storage to loop over the call to list, rather than changing how list works. Since the problem is really that StreamDir doesn't loop over list.

I could go either way. I'm not sure whose "job" it is to make sure that the requested count is served back up the pipeline.
But I do think that moving the loop up the stack like you suggest would preserve our ability to use List to get a single page. So I'm happy to make this change.

Ferelith-maker (approved):

Dabnsky (approved):

PR #31, opened 2023-10-27
Fix gEdit Bug

foodprocessor:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Buggy behavior:
In "case 2", when a file is in the local cache but not in the container, renaming the file would fail with an input/output error (no source file in container).
If the "create-empty-file" config option was set true, then the software would write an empty object to the container when the file was created. Then rename would write an empty destination file in the container (empty source file in container).

Desired behavior:
Regardless of whether "create-empty-file" is set, the updated local file data should be written to the destination in the container on rename, and no error should be returned.

Summary of changes:
When RenameFile() is called in the file_cache, if the source is local and has an open handle, upload the source directly to the destination.
Honor FSync calls by calling Sync on the local file.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues
https://github.com/Azure/azure-storage-fuse/issues/1259
https://github.com/Azure/azure-storage-fuse/issues/1060

- Related Issue #
- Closes #

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -972,6 +972,8 @@ func (fc *FileCache) SyncFile(options internal.SyncFileOptions) error {
 	if fc.syncToFlush {
 		options.Handle.Flags.Set(handlemap.HandleFlagDirty)
 	} else {
+		// TODO: shouldn't we get the local cache fd and call fsync on that here?
+		// 		 otherwise, this does basically nothing.
 		err := fc.NextComponent().SyncFile(options)
 		if err != nil {
 			log.Err("FileCache::SyncFile : %s failed", options.Handle.Path)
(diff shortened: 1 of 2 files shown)


foodprocessor:
Submitted this as a [PR to blobfuse2](https://github.com/Azure/azure-storage-fuse/pull/1277), but it was rejected.
The discussion in the PR has some ideas about how to improve this implementation.
I think this implementation is acceptable, since it is simple to read, and because upstream is likely to fix this bug soon (planned for 2.1.2), which might be released somewhere around the new year.

Ferelith-maker (approved):
I thinking if we want to pull your fix until it's fixed upstream we should have a jira task that we re-evaluate consistently.

Dabnsky:
what's the plan here? do we want to merge this in or wait for them to fix it?

foodprocessor:
I've updated the code to improve performance and follow the recommendations that came out of the PR discussion with BF2.
I think we should merge this solution now, and deal with upstream's solution incoming once that happens.
Upstream may not fix this for a long time. We should not depend on them too heavily.

Dabnsky (approved):

foodprocessor:
Upstream has fixed this (very) recently by fixing the sync-to-flush flag, so I'm going to go ahead and replace this fix with upstream's solution.
The only difference I'll introduce is to default the sync-to-flush flag to true, so that this works the way we would like by default.
I'll go ahead and do all this on this same PR, for continuity's sake.

foodprocessor:
Closing this PR to pull from upstream in a separate PR.

PR #32, opened 2023-10-27
Windows installer

jfantinhardesty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Add support to build an installer for Windows. This uses free the Inno Setup tool. The installer also identifies if the user has WinFSP installed on their computer, and if not it starts the WinFSP installer which is bundled inside of our installer. It also installs cloudfuse as a Windows service and starts the service.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [x] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


build/winfsp-2.0.23075.msi
foodprocessor, line 1: Is committing the binary to the repo best practice? Is there a way to avoid this (e.g. automatically, or even manually fetching it)?

jfantinhardesty: That is possible. I thought that keeping it in our source code makes it clear when we upgrade WinFsp versions and keeps the build process a bit simpler. Right now this is a very manual process to build the Windows installer and I was trying to reduce all of the manual parts as much as possible.

foodprocessor: That makes sense. I'm just wary of permanently increasing the cost of git clone.
Could we commit the URL to the version of WinFSP we're using instead? Maybe a script that downloads it?
The WinFSP download URLs seem like they would lend themselves to that well (the current one is "https://github.com/winfsp/winfsp/releases/download/v2.0/winfsp-2.0.23075.msi").

Ferelith-maker (approved):

build/windows_installer_build.iss
Dabnsky, line 1: this .iss file is so alien to me. what resources or tutorials have you used for reference to learn how to do this?

jfantinhardesty: It's just a file type for this specific software Inno Setup. I just looked through some of their documentation and they have a starting file to get some of it started. https://jrsoftware.org/isinfo.php

foodprocessor (approved):
🥳

PR #44, opened 2023-11-14
GUI change default config file directory

Ferelith-maker:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Change the default directory in windows to appdata/local/Cloudfuse to avoid needing admin rights to modify the config file. The GUI assumes the directory already exists.
### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


gui/common_qt_functions.py
```diff
@@ -1,11 +1,16 @@
...
+# Import QT libraries
 from PySide6 import QtWidgets
 from PySide6.QtWidgets import QWidget
 from PySide6.QtCore import Qt, QSettings
-import yaml
-import os
 
 file_cache_eviction_choices = ['lru','lfu']
 libfusePermissions = [0o777,0o666,0o644,0o444]
+defaultFuseDir = '\Cloudfuse'
```
foodprocessor, line 13: Can you drop the slash here, and just add it when you're building the path?
You could use os.path.join() to build paths without having to hard-code any slashes, which may become important if this variable gets used in Linux at some point.

gui/mountPrimaryWindow.py
```diff
@@ -18,6 +18,8 @@
 
 bucketOptions = ['s3storage', 'azstorage']
 mountTargetComponent = 3
+defaultFuseDir = '\Cloudfuse'
```
foodprocessor, line 21: Since this is already defined in common_qt_functions.py, is it possible to define this in just one place, with other shared constants?

gui/mountPrimaryWindow.py
```diff
@@ -211,7 +213,13 @@ def unmountBucket(self):
...
     # This function reads in the config file, modifies the components section, then writes the config file back
     def modifyPipeline(self,target):
-        currentDir = os.getcwd()
+
+        if platform == "win32":
+            userDir = os.getenv('LOCALAPPDATA')
+            currentDir = userDir + defaultFuseDir
+        else:
+            currentDir = os.getcwd()
+
```
foodprocessor, line 222: This code is identical to getCurrentDir(). Can you call that here instead?

gui/common_qt_functions.py
foodprocessor, line 13: These variables are only used inside widgetCustomFunctions. Should they be member fields of that class?

jfantinhardesty (approved):

gui/common_qt_functions.py
Ferelith-maker: I put it in the function itself

Ferelith-maker: Done

gui/mountPrimaryWindow.py
Ferelith-maker: Agreed

PR #46, opened 2023-11-16
Streaming changes blobfuse2

jfantinhardesty:

- [] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This pulls in new changes to blobfuse2 to fix issues with data loss on streaming mode.

### Checklist

- [x] Tested locally
- [x] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #


component/azstorage/azauthmsi.go
```diff
@@ -52,6 +53,11 @@ type azAuthMSI struct {
 	azAuthBase
 }
 
+func getNextExpiryTimer(token *adal.Token) time.Duration {
+	delay := time.Duration(5+rand.Intn(120)) * time.Second
```
Dabnsky, line 57: I'm very intrigued on why we want an inconsistent time duration interval.

jfantinhardesty: Yeah there was in issue with blobfuse2 where certain customers had many concurrent mounts and the time delays were all the same. So requests for authorization would be sent at the same time, and some requests would be denied because they had too many requests at the same time. So this prevents all of those requests from being sent at the same time.

Ferelith-maker (approved):

Dabnsky (approved):

foodprocessor:
blobfuse2 released version 2.1.2 yesterday. It's exactly the version you merged here, plus a CHANGELOG diff.
Since your history is pretty clean, it seems no harm to just merge their 2.1.2 into this PR.

component/azstorage/block_blob_test.go
```diff
@@ -3375,6 +3418,85 @@ func (s *blockBlobTestSuite) TestInvalidMD5OnReadNoVaildate() {
 // 	s.az.CloseFile(internal.CloseFileOptions{Handle: h})
 // }
 
+func (suite *blockBlobTestSuite) TestTruncateSmallFileToSmaller() {
+	suite.UtilityFunctionTestTruncateFileToSmaller(2*MB, 1*MB)
```
foodprocessor, line 3422: I see you made these file sizes significantly smaller than the ones used by BF2.
I assume the intent here is to reduce our cloud storage costs for testing?

component/azstorage/block_blob_test.go
```diff
@@ -3375,6 +3418,85 @@ func (s *blockBlobTestSuite) TestInvalidMD5OnReadNoVaildate() {
...
+}
+
+func (suite *blockBlobTestSuite) TestTruncateNoBlockFileToLarger() {
+	suite.UtilityFunctionTruncateFileToLarger(10*MB, 20*MB)
+}
+
+func (suite *blockBlobTestSuite) UtilityFunctionTestTruncateFileToSmaller(size int, truncatedLength int) {
+	defer suite.cleanupTest()
+	// Setup
+	vdConfig := fmt.Sprintf("azstorage:\n  account-name: %s\n  endpoint: %s\n  type: block\n  account-key: %s\n  mode: key\n  container: %s\n  fail-unsupported-op: true\n  virtual-directory: true",
```
foodprocessor, line 3444: You have endpoint defined directly here, but their version uses blob storage:
...`endpoint: https://%s.blob.core.windows.net/\n`...
This is so we can use Azurite in testing, right?
So we always remove their default endpoint in test files, but leave it intact for normal operation in case a user wants to access their Azure Blob Storage container by BlockAccount.
Am I understanding that all correctly?

foodprocessor (approved):
Please merge in blobfuse2-2.1.2 and update CHANGELOG.md before merging.
This looks good!

component/azstorage/block_blob_test.go
jfantinhardesty: Yes, that was the intention here.

component/azstorage/block_blob_test.go
jfantinhardesty: Yes, this is so that it can connect to both Azurite and a regular blob endpoint for testing, since our tests do both. And I just change how that endpoint is specified for testing files. This doesn't change how the endpoint is used by users when running cloudfuse.

PR #48, opened 2023-11-17
Bugfix: Access Denied on Windows When Running as a Service

foodprocessor:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

We were giving file ownership to the "current user" (-1), but when launching as a service, the current user is SYSTEM, which does not allow files to be modified by any users (including admin!).
This changes ownership to the "Authenticated Users" group, which allows writes but stops short of allowing writes for the EVERYONE group.
This bug, and the unsophisticated fix this PR introduces suggest that we need to give more serious treatment to security / permissions on Windows at some future point.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/libfuse/libfuse2_handler.go
+++ b/component/libfuse/libfuse2_handler.go
@@ -101,11 +101,11 @@ func (lf *Libfuse) initFuse() error {
 		lf.attributeExpiration,
 		lf.negativeTimeout)
 
-	// With WinFSP this will present all files as owned by the current user and group
+	// With WinFSP this will present all files as owned by the Authenticated Users group
 	if runtime.GOOS == "windows" {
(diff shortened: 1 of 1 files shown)


jfantinhardesty (approved):
I think is a good quick fix. Based on this https://github.com/winfsp/winfsp/wiki/Frequently-Asked-Questions there is a tool called fsptool that comes with WinFSP that can convert between Windows accounts and uid's, so maybe we could use this in the future to identify what the current user is when mounting.

PR #57, opened 2023-12-04
added "Make sure that WinFSP is installed" to error

Dabnsky:
adds verbiage to GUI error window to check WinFSP being installed


foodprocessor (commented):
Could you add a messageBox to the GUI to tell the user to install WinFSP when this error comes back?
That would be in gui/mountPrimaryWindow.py: mountBucket()

foodprocessor:
I just spotted a bug that I wrote on [a previous PR](https://github.com/Seagate/cloudfuse/pull/58), which fails to return an error from `GetCurrentUser` when WinFSP is not installed.
I'd like to push the fix for that to this PR.

cmd/service_windows.go
```diff
@@ -124,7 +124,7 @@ var mountServiceCmd = &cobra.Command{
 
 		err = mountInstance()
 		if err != nil {
-			return fmt.Errorf("failed to mount instance [%s]", err.Error())
+			return fmt.Errorf("failed to mount instance [%s]\n Verify WinFSP is installed correctly", err.Error())
```
jfantinhardesty, line 127: Does this error only occur if WinFSP is not installed? Or could there be another reason. I would just want to try and isolate the error for WinFSP so that users don't feel frustrated needing to verify their WinFSP installation when the issue could be something else.

jfantinhardesty (changes requested):

cmd/service_windows.go
Dabnsky: This error is added in place without checking if WinFSP is already installed. I see Michael's comment about the push he put in place on where the specific error can be output. I'll use his implementation

jfantinhardesty: Yeah I think that fix would work well.

Dabnsky:
Michael's bug fix addressed the changes I was attempting to make. This PR is no longer relevant

foodprocessor:
Because this PR never merged, my bugfix was never merged either...

PR #58, opened 2023-12-07
Fix Windows Permissions - Pass Current User to WinFSP

foodprocessor:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

When mounting using the WinFSP service on Windows, pass the current user uid and gid to the cloudfuse service instance.
This way the owner uid and gid are set to the current user, instead of trying to find a user-agnostic owner (which was compromising security).
No extra CLI options were needed, since uid and gid can already be passed in using the -o CLI switch.

Some of the changes on this PR are not related to permissions at all:

- build/windows_installer_build.iss:
Properly detect whether WinFSP is installed, and don't prompt for unnecessary WinFSP installation
- cmd/mount.go:
Don't describe the mount command as Azure-specific
- component/libfuse/libfuse2_handler.go:96
Set libfuse option to reduce extra GetAttr calls on Windows

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


common/util.go
```diff
@@ -120,6 +121,26 @@ func GetCurrentUser() (uint32, uint32, error) {
 		return 0, 0, err
 	}
 
+	if runtime.GOOS == "windows" {
```
jfantinhardesty, line 124: I am not sure that this is needed anymore with my latest pull request. Since we don't have the Windows service launching the programs, then WinFSP should always know who the current user.

jfantinhardesty (changes requested):

common/util.go
foodprocessor: From our discussions, it sounds like this is the easiest path forward.
I'll leave this be and resolve this.

jfantinhardesty (approved):

PR #59, opened 2023-12-08
Windows mount without admin

jfantinhardesty:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

In this pull request we no longer use the Windows registry to track existing mounts on Windows. These are now tracked in a 
mounts.json file in the users AppData folder.

We no longer use a Windows service to run in the background to restart mounts on system reboot, this is now taken care of by a windows-startup tool that is installed for the user to run as a startup program which will relaunch mounts that are in the mounts.json file.  This now means that we no longer need to run the GUI as admin or use the `service mount` command as admin. The Windows installer from Inno Setup was also changed a bit to correctly set Cloudfuse in the path, which was broken before and will install the windows-startup tool for the user.

I also fixed the Windows installer and Go releaser to include the health-monitor binary which was being excluded from our builds previously.

I believe that I updated most of the documentation to account for these changes, but please let me know if there is something that I missed.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [x] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


internal/winservice/service_windows.go
```diff
@@ -142,13 +96,11 @@ func IsMounted(mountPath string) (bool, error) {
 }
 
 // startService starts cloudfuse by instructing WinFsp to launch it.
-func startServices() error {
+func StartMounts() error {
 	// Read registry to get names of the instances we need to start
-	instances, err := readRegistryEntry()
+	instances, err := readMountsFromFile()
 	// If there is nothing in our registry to mount then continue
```
foodprocessor, line 102: These comments need updating ("service" and "registry") are obsolete here.

foodprocessor (approved):
Looks good!

Ferelith-maker (approved):

PR #60, opened 2023-12-12
Keep cache entries valid

foodprocessor:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

Reorganized attr_cache a bit.
Now keeping cache entries valid whenever possible.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


component/attr_cache/attr_cache_test.go
```diff
@@ -898,15 +898,22 @@ func (suite *attrCacheTestSuite) TestCreateFile() {
 
 	_, err = suite.attrCache.CreateFile(options)
 	suite.assert.Nil(err)
-	assertNotInCache(suite.assert, suite.attrCache, path)
+	assertExists(suite, options.Name)
```
jfantinhardesty, line 901: Why did we need to change this? Was createFile not creating an entry in the cache before?

jfantinhardesty (changes requested):

component/attr_cache/attr_cache.go
```diff
@@ -796,21 +796,21 @@ func (ac *AttrCache) GetAttr(options internal.GetAttrOptions) (*internal.ObjAttr
 	return pathAttr, getErr
 }
 
-// CreateLink : Mark the link and target invalid
+// CreateLink : Mark the new link invalid
```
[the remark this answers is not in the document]
foodprocessor: Good catch!
I just fixed this.
I also couldn't resist, and fixed all the asserts in the test file to use receivers (`suite.assertBlah(path, etc)` instead of `assertBlah(suite, path, etc)`.

component/attr_cache/attr_cache.go
```diff
@@ -821,14 +821,7 @@ func (ac *AttrCache) CreateLink(options internal.CreateLinkOptions) error {
...
 	err := ac.NextComponent().FlushFile(options)
-	if err == nil {
-		ac.cacheLock.RLock()
-		defer ac.cacheLock.RUnlock()
-		toBeInvalid, getErr := ac.cacheMap.get(options.Handle.Path)
-		if getErr == nil {
-			toBeInvalid.invalidate()
-		}
-	}
+	// we already updated the cache when WriteFile was called, so don't invalidate the entry
```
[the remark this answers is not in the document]
foodprocessor: Good question! Heh...
I thought `FlushFile` was called after multiple calls to `WriteFile`... but I'm realizing that the `Write` calls stop at the `file_cache`, then `FlushFile` is called when the file closes. So this really does need to be invalidated!

component/attr_cache/attr_cache_test.go
foodprocessor: Correct. Previously, `CreateFile` would invalidate the attributes for the file, since that's what they would do any time the file changed in cloud storage.

foodprocessor (commented):
Responding to all feedback

component/attr_cache/attr_cache.go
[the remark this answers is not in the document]
jfantinhardesty: So this needs to be changed back to how it was before then?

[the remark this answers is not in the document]
foodprocessor: Yes. I pushed a revert for this. Not sure why it's not updating here in the conversation view.

foodprocessor:
It looks like this change somehow breaks chmod (failed in code coverage). I'm not sure why, and I'm not going to merge it until I figure it out.

component/attr_cache/attr_cache.go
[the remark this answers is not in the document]
jfantinhardesty: Oh ok, yeah now I see it.

Ferelith-maker (approved):

component/attr_cache/attr_cache.go
```diff
@@ -556,11 +550,13 @@ func (ac *AttrCache) DeleteFile(options internal.DeleteFileOptions) error {
...
 		defer ac.cacheLock.RUnlock()
 		toBeDeleted, getErr := ac.cacheMap.get(options.Name)
-		if getErr != nil {
-			log.Err("AttrCache::DeleteFile : %s", getErr)
-		} else {
-			toBeDeleted.markDeleted(deletionTime)
+		if getErr != nil || !toBeDeleted.valid() {
+			log.Warn("AttrCache::DeleteFile : %s no valid entry found. Adding entry...", options.Name)
+			// add deleted file entry
+			attr := internal.CreateObjAttr(options.Name, 0, deletionTime)
```
Dabnsky, line 556: What are the benefits of adding the item to the cache right before marking it as deleted?

component/attr_cache/attr_cache.go
```diff
@@ -501,46 +478,63 @@ func (ac *AttrCache) RenameDir(options internal.RenameDirOptions) error {
...
 			// record that the parent directory tree contains at least one object
-			ac.markAncestorsInCloud(getParentDir(options.Name), time.Now())
+			ac.markAncestorsInCloud(getParentDir(options.Name), currentTime)
 		}
-		// TODO: we assume that the OS will call GetAttr after this.
-		// 		if it doesn't, will invalidating this entry cause problems?
-		toBeInvalid, getErr := ac.cacheMap.get(options.Name)
-		if getErr == nil {
-			toBeInvalid.invalidate()
+		newFileEntry, getErr := ac.cacheMap.get(options.Name)
```
jfantinhardesty, line 530: Another thought here, so before they would check that if there was an entry existing for the file they invalidate it, wheras we create it. Do we need to worry about an old cache item being in the cacheMap or can we be guaranteed that won't happen or won't cause an issue?

component/attr_cache/attr_cache.go
foodprocessor: The attribute cache needs to keep a record of the fact that this file is deleted.
I thought about passing `false` for exists here, but it seemed slightly more readable to:

1. Find or create the entry
2. Mark it deleted

component/attr_cache/attr_cache.go
```diff
@@ -315,18 +292,18 @@ func (ac *AttrCache) CreateDir(options internal.CreateDirOptions) error {
...
+			if getErr == nil && dirAttrCacheItem.exists() {
 				return os.ErrExist
 			}
-			newDirAttr := internal.CreateObjAttrDir(options.Name)
-			newDirAttrCacheItem := ac.cacheMap.insert(newDirAttr, true, time.Now())
+		}
+		// add the directory to the cache
+		newDirAttr := internal.CreateObjAttrDir(options.Name)
+		newDirAttrCacheItem := ac.cacheMap.insert(newDirAttr, true, time.Now())
+		if ac.cacheDirs {
```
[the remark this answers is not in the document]
foodprocessor: Although I agree that two `if` blocks with the same condition near each other should usually be consolidated, in this case, there's a sequence of logic:

1. If `ac.cacheDirs` is set, then the attribute cache is the authority on whether the directory to be created already exists. So if it does, return an error and do not create the folder.
2. Regardless of whether `ac.cacheDirs` is set, we want to have record of this new directory in the attribute cache, so we add that without any conditional.
3. At this point, if `ac.cacheDirs` is set, we need to mark this new directory in the cache as not existing in the cloud.

Consolidating the if blocks here would mean the code to create the new directory entry would be duplicated (for `ac.cacheDirs` true and false).
I figured that duplicating the short conditional would repeat less code, and be more readable.

component/attr_cache/attr_cache.go
```diff
@@ -315,18 +292,18 @@ func (ac *AttrCache) CreateDir(options internal.CreateDirOptions) error {
 		ac.cacheLock.Lock()
 		defer ac.cacheLock.Unlock()
 		if ac.cacheDirs {
+			// the cloud is not directory-aware
 			// check if directory already exists
-			if ac.pathExistsInCache(options.Name) {
+			dirAttrCacheItem, getErr := ac.cacheMap.get(options.Name)
```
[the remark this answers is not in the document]
foodprocessor: The difference is the error being returned.
We can check if the directory already exists, but if `ac.cacheDirs` is false, we're not supposed to return an error. Because then the cloud is the authority on whether this operation succeeded.
So if `ac.cacheDirs` is false, we don't care if the directory already exists. It's being replaced.

component/attr_cache/attr_cache.go
foodprocessor: Since our `insert` function uses a map for each directory's files, insert will replace any entry with the same path.
Here the code checks if the file exists, and if it does, it updates its size, mode, and timestamp.
As I'm re-reading this code, I think it would have been better to just replace the entry outright, and not update an old entry.
I'll add a separate suggestion comment here shortly for your review.

component/attr_cache/attr_cache.go
```diff
@@ -501,46 +478,63 @@ func (ac *AttrCache) RenameDir(options internal.RenameDirOptions) error {
...
-		if getErr == nil {
-			toBeInvalid.invalidate()
+		newFileEntry, getErr := ac.cacheMap.get(options.Name)
+		if getErr != nil || !newFileEntry.exists() {
+			// add new entry
+			newFileAttr := internal.CreateObjAttr(options.Name, 0, currentTime)
+			newFileEntry = ac.cacheMap.insert(newFileAttr, true, currentTime)
 		}
+		newFileEntry.setSize(0, currentTime)
+		newFileEntry.setMode(options.Mode)
```
foodprocessor, line 537:
```suggestion
		// add new entry
		newFileAttr := internal.CreateObjAttr(options.Name, 0, currentTime)
		newFileEntry = ac.cacheMap.insert(newFileAttr, true, currentTime)
		newFileEntry.setMode(options.Mode)
```
I think this is a simpler and better implementation.
No re-using a left-over entry which should have been invalidated, and since `insert` replaces old entries, we don't even need to look up the old entry to invalidate it.

jfantinhardesty: Do we need to call setSize here, or was that unnecessary.

jfantinhardesty (approved):

component/attr_cache/attr_cache.go
foodprocessor: The size is set by `CreateObjAttr` (the `0` parameter).

PR #63, opened 2023-12-14
Dry run

foodprocessor:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Add CLI option to do a dry run to validate config, credentials, etc.
Use dry-run in GUI to test config before mounting.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [x] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


cmd/mount.go
```diff
@@ -419,6 +421,12 @@ var mountCmd = &cobra.Command{
 			return Destroy(fmt.Sprintf("failed to initialize new pipeline [%s]", err.Error()))
 		}
 
+		// Dry run ends here
+		if options.DryRun {
+			log.Crit("Dry-run complete")
```
jfantinhardesty, line 426: Why is this logging as critical, instead of as a log.trace? It seems like this could confuse a user looking at the log file.

jfantinhardesty (approved):

cmd/mount.go
foodprocessor: I had trouble getting this message to show up in the log. But honestly, Crit didn't actually help. Something's weird about trying to log something when you're just about to exit.
Anyway, I'll switch this to trace.

gui/mountPrimaryWindow.py
```diff
@@ -114,6 +115,23 @@ def mountBucket(self):
                 self.errorMessageBox(f"Error: Cloudfuse needs to create the directory {directory}, but it already exists!")
                 return
             
+            # do a dry run to validate options and credentials
+            commandParts = ['cloudfuse.exe', 'mount', directory, f'--config-file={configPath}', '--dry-run']
```
Ferelith-maker, line 119: Is a button needed to do a dry-run, or is the intent to do a dry-run every time the mount button is pressed?

foodprocessor: Here, the intent is to run it every time, as a validation step before mounting. This should eliminate a bunch of silent failures.
If we want to have a button that does this, that's definitely an option. I think it would be nice for the GUI to have a "check" button in the config window, for example, but that's not urgent.

Dabnsky (approved):

Ferelith-maker (approved):

PR #69, opened 2023-12-19
Add and check config-version

foodprocessor:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Added config-version to config options and set version to 1.
I did not change the GUI to handle this new option, mostly because I don't know how to do that properly.
Maybe we can address that in a future PR.

### Checklist

- [ ] Tested locally
- [ ] Added new dependencies
- [x] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -64,6 +64,7 @@ type mountOptions struct {
 	ConfigFile string
 
 	DryRun            bool
+	ConfigVersion     int            `config:"config-version"`
 	Logging           LogOptions     `config:"logging"`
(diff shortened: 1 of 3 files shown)


foodprocessor:
If at some point we change the config file options enough to break backward compatibility, we should introduce the version number then. At that point we can warn users that omit the version number that their config file is being interpreted as a down-rev (version 1) config file.

PR #83, opened 2024-01-05
List mount points on windows

Dabnsky:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Added windows support for the ListMountPoints() in the common.util.go

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


common/util.go
```diff
@@ -176,7 +176,21 @@ func ListMountPoints() ([]string, error) {
 	// TODO: Add support to list current mounts
 	// We cannot list mount points like we do on Linux
 	if runtime.GOOS == "windows" {
-		return nil, nil
+
+		out, err := exec.Command(`C:\Program Files (x86)\WinFsp\bin\launchctl-x64.exe`, "list").Output()
+		if err != nil {
+			fmt.Printf("Is WinFSP installed? 'launchctl-x64.exe list' failed with error: %v\n", err)
+		}
```
foodprocessor, line 183: Did you mean to return `err` here?

common/util.go
```diff
@@ -176,7 +176,21 @@ func ListMountPoints() ([]string, error) {
 	// TODO: Add support to list current mounts
 	// We cannot list mount points like we do on Linux
 	if runtime.GOOS == "windows" {
```
foodprocessor, line 178: This function seems to break pretty neatly by platform (no code is shared), so it might be nice to move this whole function out of this file, and instead put the two implementations into separate platform-specific files: util_windows.go and util_linux.go.

foodprocessor (changes requested):
Looks pretty good!
Does this mean `./cloudfuse.exe mount list` works on Windows now?

common/util.go
```diff
@@ -176,7 +176,21 @@ func ListMountPoints() ([]string, error) {
 	// TODO: Add support to list current mounts
```
jfantinhardesty, line 176: Could you remove this todo here.

common/util.go
Dabnsky: I pushed a return nil, err for this one

common/util.go
Dabnsky: The other function, GetCurrentUser(), would need to be moved for this as well. I would consider moving these to a separate windows file as a separate task for the backlog. what do you think?

common/util.go
Dabnsky: done

common/util.go
foodprocessor: You could split `GetCurrentUser` into two separate implementations, but there are at least a few lines of code shared by the two, which makes a case for leaving it the way it is.
In the case of ListMountPoints, nothing is shared.
Moving the functions to separate files could be another PR, but I think if we're going to do it, we should just go ahead and do it in this PR. It's not a big lift.

Dabnsky: this was a misunderstanding of having to create linux and windows files and split up all function windows sections into the windows file. util_windows.go already exists and I'll just move the windows portion of the function to that file.

foodprocessor (approved):
✔

PR #92, opened 2024-01-12
Friendly GUI Default Settings

foodprocessor:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Default settings in GUI and in sample config files to allow for a persistent file cache.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/gui/common_qt_functions.py
+++ b/gui/common_qt_functions.py
@@ -105,23 +105,24 @@ class defaultSettingsManager():
         self.settings.setValue('file_cache',{
             'path': '',
             'policy': 'lru',
-            'timeout-sec' : 120,
+            'timeout-sec' : 64000000,
             'max-eviction': 5000,
             'max-size-mb': 0,
             'high-threshold': 80,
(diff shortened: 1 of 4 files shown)


jfantinhardesty (approved):
Looks good. While looking at these I was also wondering if we should set "create-empty-file: true" so that they don't run into the issues we had previously observed with some text editors.

foodprocessor:
> Looks good. While looking at these I was also wondering if we should set "create-empty-file: true" so that they don't run into the issues we had previously observed with some text editors.

sync-to-flush is true by default now, so this shouldn't be an issue.
If they do disable sync-to-flush, then they will lose data in that case - either with a silent failure (create-empty-file: true), or with an I/O error (create-empty-file: false).
I slightly prefer the noisy failure to the silent one.

PR #97, opened 2024-01-17
GUI Config Validation

foodprocessor:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Make GUI enforce config YAML schame based on default values.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/gui/common_qt_functions.py
+++ b/gui/common_qt_functions.py
@@ -284,7 +284,8 @@ class widgetCustomFunctions(QWidget):
                     tempDict[suboption] = dictForConfigs[option][suboption]
                 self.settings.setValue(option,tempDict)
             else:
-                self.settings.setValue(option,dictForConfigs[option])
+                if dictForConfigs[option] is not None:
+                    self.settings.setValue(option,dictForConfigs[option])
 
     # Check for a true/false setting and set the checkbox state as appropriate. 
     #   Note, Checked/UnChecked are NOT True/False data types, hence the need to check what the values are.
--- a/gui/common_qt_functions.py
+++ b/gui/common_qt_functions.py
@@ -278,14 +278,19 @@ class widgetCustomFunctions(QWidget):
     def initSettingsFromConfig(self):
         dictForConfigs = self.getConfigs()
         for option in dictForConfigs:
+            # check default settings to enforce YAML schema
+            invalidOption = not self.settings.contains(option)
+            invalidType = type(self.settings.value(option)) != type(dictForConfigs[option])
+            if invalidOption or invalidType:
+                print(f"WARNING: Ignoring invalid config option: {option} (type mismatch)")
+                continue
             if type(dictForConfigs[option]) == dict:
                 tempDict = self.settings.value(option)
                 for suboption in dictForConfigs[option]:
                     tempDict[suboption] = dictForConfigs[option][suboption]
                 self.settings.setValue(option,tempDict)
             else:
-                if dictForConfigs[option] is not None:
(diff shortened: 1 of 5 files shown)


gui/common_qt_functions.py
```diff
@@ -69,21 +78,30 @@ def setAzureSettings(self):
             'auth-resource': '',
             'update-md5': False,
             'validate-md5': False,
-            'virtual-directory': False,
```
Ferelith-maker, line 72: I thought this was causing issues and we set it to False for a reason, did we fix this so it can be true?

jfantinhardesty (approved):

gui/common_qt_functions.py
```diff
@@ -69,21 +78,30 @@ def setAzureSettings(self):
...
-            'disable-compression': False
+            'virtual-directory': True,
+            'disable-compression': False,
+            'max-results-for-list': 2,
```
Ferelith-maker, line 83: I think we talked about this already, so I wanted to add this in the comment for future reference. These new settings will not be expressed by the GUI until the .ui code is modified to show these. I am fine using the code here to enforce a yaml schema, we will just need to note what changed in this PR to then update the ui code later.

gui/common_qt_functions.py
```diff
@@ -102,6 +121,16 @@ def setComponentSettings(self):
             'buffer-size-mb': 0,
             'file-caching': False # false = handle level caching ON
         })
+        self.settings.setValue('block_cache',{
```
Ferelith-maker, line 124: Adding block_cache might be enough of a major change to justify adding components to the GUI page to reflect this.

gui/common_qt_functions.py
foodprocessor: I'll add some comments here to indicate the options that are not exposed in the GUI itself.
I think it's okay to leave many options invisible to GUI users. Users can always edit the config file directly.

Ferelith-maker (approved):

gui/common_qt_functions.py
foodprocessor: We can definitely add GUI components for the block cache, but I think that can be a separate PR.
There may be an issue with the GUI removing block_cache from a pipeline written by the user.
I've added a story for these questions.
For now, I'd like to disjoint the GUI's YAML validation from its actual settings interface. I've added a story for re-evaluating which options should be exposed.

gui/common_qt_functions.py
foodprocessor: I think this setting was just flagged false because most booleans are false by default. This one is true by default. And I think setting it true actually makes the code more compatible (it allows blobs to show up whose parent folders don't have a folder marker in cloud storage).

gui/common_qt_functions.py
foodprocessor: I added `block_cache` to the dropdown menu in S3 and Azure settings, and added it to the list of possible components.
I did a quick run test, and it works (the drop-down shows up, it doesn't clobber the config, and it loads from the config correctly).

gui/common_qt_functions.py
Ferelith-maker: Agreed.

gui/common_qt_functions.py
Ferelith-maker: Nice job

Ferelith-maker (approved):
Nicely done

Issue #81, opened 2024-01-05
Listing container contents is often slow


foodprocessor:
### Problem

I have a relatively large number of files (about 2,000) in the root directory in my container. When I first mount the container, listing the root directory sometimes takes an excessive amount of time. At the moment it's 5s. But it's been over 20s at times.

### Expected behavior

Listing 2,000 files should take much less time. I think 1s is ideal, 5s is reasonable, and 10s is not.

### How to reproduce

List a directory with a few thousand files right after mounting:
```
cloudfuse mount ~/mycontainer --config-file=config.yaml
time ls ~/mycontainer
cloudfuse unmount --lazy ~/mycontainer
```

### Logs

[cloudfuse_list_2k.log](https://github.com/Seagate/cloudfuse/files/13844978/cloudfuse_list_2k.log)

### Deployment information

I'm on Ubuntu 22.04, running against a Seagate Lyve Cloud container, running cloudfuse v0.3.0.

foodprocessor:
It looks like the issue that came up in your log is just that the OS called Getattr on every single entry in the directory.
It seems this is because we use fuse2 and not fuse3 for our fuse connector, and fuse2 lacks a feature that prevents the OS from calling Getattr on all entries in each folder it lists.
The reason we're using fuse2 is that we use Cgofuse for our fuse layer, which [doesn't support fuse3 yet](https://github.com/winfsp/cgofuse/issues/50).
It looks like the performance penalty is roughly 1s for every 500 entries.

PR #107, opened 2024-01-29
Cache Directory List

foodprocessor:
Bugfix: bad math in S3storage::StreamDir.
Keep filling dir list cache when OS re-lists from the beginning (>MaxDirListCount entries).

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [x] Optimization
- [ ] Documentation Update

Use attr_cache to respond to StreamDir and ReadDir without using cloud storage.
Respect attr_cache `timeout-sec`.
Use repeated OS `Readdir` calls with zero offset to complete directory listing despite bad OS behavior.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Closes [#81](https://github.com/Seagate/cloudfuse/issues/81)

### DIFF
--- a/component/attr_cache/attr_cache.go
+++ b/component/attr_cache/attr_cache.go
@@ -487,6 +487,10 @@ func (ac *AttrCache) fetchCachedDirList(path string, token string) ([]*internal.
 	log.Trace("AttrCache::fetchCachedDirList : %s token=\"%s\"", path, token)
 
 	var pathList []*internal.ObjAttr
+	if !ac.cacheOnList {
+		log.Debug("AttrCache::fetchCachedDirList : %s cache on list is disabled", path)
+		return pathList, "", fmt.Errorf("cache on list is disabled")
+	}
 	listDirCache, found := ac.cache.get(path)
 	if !found {
 		log.Debug("AttrCache::fetchCachedDirList : %s directory not found in cache", path)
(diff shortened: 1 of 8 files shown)


component/attr_cache/attr_cache.go
```diff
@@ -354,10 +374,17 @@ func (ac *AttrCache) ReadDir(options internal.ReadDirOptions) (pathList []*inter
...
 			pathList, numAdded = ac.addDirsNotInCloudToListing(options.Name, pathList)
-			log.Trace("AttrCache::ReadDir : %s +%d from cache = %d",
+			log.Info("AttrCache::ReadDir : %s +%d from cache = %d",
 				options.Name, numAdded, len(pathList))
 		}
 	}
+
+	// values should be returned in ascending order by key, without duplicates
+	// sort
+	slices.SortFunc[[]*internal.ObjAttr, *internal.ObjAttr](pathList, func(a, b *internal.ObjAttr) int {
```
jfantinhardesty, line 384: Nice use of the slices package!

jfantinhardesty (approved):

jfantinhardesty:
One other thought I had is we should add documentation somewhere that the caching directories could cause an issue if something other than the current instance of Cloudfuse is modifying the bucket, that these won't be observed locally until the timeout expires. We don't guarantee multitenancy, but these changes do affect that a bit so I think we should at least document that.

component/attr_cache/attr_cache.go
foodprocessor: I read that this is faster than `sort.Sort`, and I was using this difficult generic thing for removing duplicates below, so I figured why not! 😅

component/attr_cache/attr_cache.go
```diff
@@ -403,26 +441,84 @@ func (ac *AttrCache) StreamDir(options internal.StreamDirOptions) ([]*internal.O
...
+		return pathList, "", fmt.Errorf("%s directory not found in cache", path)
+	}
+	log.Debug("AttrCache::fetchCachedDirList : %s listing token=\"%s\"", path, listDirCache.listToken)
+	// check timeout
+	if time.Since(listDirCache.listedAt).Seconds() >= float64(ac.cacheTimeout) {
+		log.Debug("AttrCache::fetchCachedDirList : %s listing cache expired", path)
+		return pathList, "", fmt.Errorf("%s directory listing expired", path)
+	}
+	// don't provide cached data when new (uncached) data is being requested
+	if token != "" && token == listDirCache.listToken {
```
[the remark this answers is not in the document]
foodprocessor: It's a bit confusing, honestly.
Yes, I cache the returned items and record the token that came back. But since the token that came back is meant to be used to get more entries _**from the cloud**_, when a subsequent request has a matching token, I **do not** serve from cache but route that request to the cloud.

foodprocessor:
> One other thought I had is we should add documentation somewhere that the caching directories could cause an issue if something other than the current instance of Cloudfuse is modifying the bucket, that these won't be observed locally until the timeout expires. We don't guarantee multitenancy, but these changes do affect that a bit so I think we should at least document that.

@jfantinhardesty you're right. I'm just not sure where to put that information... I'll go ahead and put it in baseConfig.yaml under cache-on-list. Let me know if you think it should be mentioned somewhere else (as well).

PR #109, opened 2024-02-01
New location linux config for gui

Ferelith-maker:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

The GUI will now create its own directory in either APPDATA in windows, or /home/.../.cloudfuse on Linux. Refactor the code to make more sense on naming and how the config file is modified in the main window
### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


gui/common_qt_functions.py
```diff
@@ -264,19 +264,20 @@ def updateSettingsFromUIChoices(self):
...
+    def getWorkingDir(self):
         if platform == "win32":
+            defaultFuseDir = 'Cloudfuse'
             userDir = os.getenv('APPDATA')
-            currentDir = os.path.join(userDir, defaultFuseDir)
         else:
-            currentDir = os.getcwd()
+            defaultFuseDir = '.Cloudfuse'
+            userDir = os.getenv('HOME')
+        currentDir = os.path.join(userDir, defaultFuseDir)
```
foodprocessor, line 274: Looks like a few more variable names could change to fit with the `currentDir` -> `workingDir` change.

gui/mountPrimaryWindow.py
```diff
@@ -250,31 +257,28 @@ def modifyPipeline(self,target):
...
         if components != None:
-            components[mountTargetComponent] = target
+            components[mountTargetComponent] = targetBucket
             configs['components'] = components
         else:
             self.errorMessageBox(
                 f"The components is missing in {currentDir}/config.yaml. Consider Going through the settings to create one.",
                 "Components in config missing")
-            return False
+            return
```
foodprocessor, line 273: Since there is not return value now, is there any risk of a failure in `modifyPipeline` causing issues?

foodprocessor (approved):
Looks good! Thanks!

jfantinhardesty (approved):

gui/mountPrimaryWindow.py
Ferelith-maker: It acts the same way the code did before, it just notifies the user that something went wrong and couldn't modify the file. There is a chance that the user might forget the pipeline wasn't modified and would try to mount the bucket anyway. This points out the need to verify the integrity of the config file before mounting, but that functionality isn't in the main page anyway. I could add that to the backlog

gui/common_qt_functions.py
Ferelith-maker: Thanks for finding those, I will change them

Dabnsky (approved):

gui/mountPrimaryWindow.py
foodprocessor: But before, `mountBucket` was gating the mount with the return value from this function.
Now if I try to mount with a config file that has an invalid `components` section, it will go ahead and try to mount.
I just tried mounting using the CLI, with components missing from the `components` section, and it just goes ahead and builds a pipeline that won't work, and give an I/O error when the OS tries to interact with the mount - a sort of silent failure on mount.
We could (and maybe should) validate the pipeline in the Cloudfuse Go code, but we don't.
I could go either way here - if it's going to break the code organization that you chose to put this validation back in, then you can leave it and make it a backlog item. But if it's easy to get it back, it would be nice to keep it.

gui/mountPrimaryWindow.py
```diff
@@ -135,21 +144,13 @@ def showUnderConstructionPage(self):
 
 
     def mountBucket(self):
-        self.addOutputText("Validating configuration...")
-        # Update the pipeline/components before mounting the target
-        targetIndex = self.dropDown_bucketSelect.currentIndex() 
-        success = self.modifyPipeline(bucketOptions[targetIndex])
-        if not success:
```
foodprocessor, line 142: This is the code I was thinking about, which gates mount with modifyPipeline.
I see that the new way you're doing it (attaching the function to the dropdown), doesn't really allow for this gate here. So I think a backlog item for that to be added, either to the GUI or the CLI, is fine.

PR #110, opened 2024-02-02
Concurrency and Coherency for Directory Listing Cache

foodprocessor:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

Change libfuse::Readdir to use fill properly with a non-zero offset
Add directory listing cache to attr_cache to serve concurrent StreamDir requests
This improves performance for Readdir by another 2x. I now see cached Readdir calls serving >8k entries in <30ms.

There is a fair amount of refactoring in this one:

- The attribute cacheMap insert function now takes an options struct
- Attribute cache items now have a handle to their parents
- Directories now have a map of listing cache segments, and each segment now has its own timestamp (and they each expire independently)
- A directory's listing cache is cleared when its contents change (except when the cache is being filled by listing from cloud storage)
- Libfuse::Readdir is broken up into a few functions for improved readability
- The issue where List returns the directory being listed but we don't want that entry is mitigated and explained
- StreamDir is called with Count=0, allowing the cloud component to set the count based on its API limits

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #

### Latest changes:

**Libfuse::Readdir:**

- Grow cacheInfo instead of replacing contents every time the pipeline is called, up to a limit of 5000 entries
- Add field cacheInfo.lastPage, which tracks whether the pipeline has any more entries to provide.  
This replaces logic that assumed there were no more entries when the token was empty and the offset was not 0.

**Attribute Cache**
- Don't clear an item's attr struct when marking it deleted or invalid
- `moveCachedItem` will no longer move deleted items
- GetAttr: When an entry does not exist, return `nil, ENOENT` instead of `&ObjAttr{}, ENOENT`

**E2E Tests**
- Print (but do not assert against) errors when cleaning up tests directories

### DIFF
--- a/component/attr_cache/cacheMap.go
+++ b/component/attr_cache/cacheMap.go
@@ -30,6 +30,7 @@ import (
 	"time"
 
 	"github.com/Seagate/cloudfuse/common"
+	"github.com/Seagate/cloudfuse/common/log"
 	"github.com/Seagate/cloudfuse/internal"
 )
 
@@ -130,7 +131,7 @@ func (ctm *cacheTreeMap) insertItem(newItem *attrCacheItem, fromDirList bool) {
 	parentPath := getParentDir(path)
 	parentItem, parentFound := ctm.get(parentPath)
 	// if there is no parent, create one and add it
-	if !parentFound || (!parentItem.exists() && newItem.exists()) {
+	if !parentFound || (!parentItem.isRoot() && !parentItem.exists() && newItem.exists()) {
 		newParentAttr := internal.CreateObjAttrDir(parentPath)
 		parentItem = newAttrCacheItem(newParentAttr, newItem.exists(), newItem.cachedAt)
 		// recurse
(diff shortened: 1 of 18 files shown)


component/attr_cache/cacheMap.go
```diff
@@ -58,6 +65,14 @@ type cacheTreeMap struct {
 	cacheTree *attrCacheItem
 }
 
+// type passed to insert
+type insertOptions struct {
```
Dabnsky, line 69: What are the main benefits to supplying options into the insert? So far it seems more readable to see what parameters are being set to what values. Does this add more flexibility to how insert can be used? if so, how?

foodprocessor: I chose to add a struct because I wanted to add a parameter (`fromDirList`) with a default value (`false`).
The struct allows me to do that, since Go does not have optional function parameters like Python does.
Another slight benefit is that filling out the `insertOptions` struct when calling `insert` makes each parameter's name visible, so there's no question why we're passing the literal `true` in as the second parameter.
The downside, of course, is how verbose it is. It seems like a lot of noise to add so many lines and words.
I'm not sure if this was the right decision, and I'm eager to hear what it's like to read this. Is it noisy? Readable? Annoying?

Dabnsky (approved):
Looks good! Thanks for the tour

jfantinhardesty (approved):

PR #158, opened 2024-03-21
Minor Improvements

foodprocessor:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

- Add version information to installer metadata
- Improve comment and log wording:
    - don't refer to cloud storage as simply "storage"
    - fix spelling errors
- Remove redundant code that sets a flag when it's already set
- Mark file as deleted when cloud storage open call fails with ENOENT
- Don't check for `IsExist` errors from `GetAttr`, `Lstat`, or `Stat`, none of which return that error
- If an attribute cache item is marked deleted, don't invalidate it and forget that just because `Chmod` was called
- Better prevent requesting a single trailing attribute entry
- Prevent file cache from attempting to delete the file it already just deleted
- Never use `&internal.ObjAttr{}`. Use `nil` instead

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


component/attr_cache/attr_cache.go
```diff
@@ -662,6 +662,27 @@ func (ac *AttrCache) CreateFile(options internal.CreateFileOptions) (*handlemap.
 	return h, err
 }
 
+// OpenFile: Update cache with Open results from cloud
+func (ac *AttrCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle, error) {
```
jfantinhardesty, line 666: Since you are adding a new function to attr_cache, could you add some unit tests for it?

component/file_cache/file_cache_test.go
```diff
@@ -1409,7 +1409,7 @@ func (suite *fileCacheTestSuite) TestStatFS() {
 
 func (suite *fileCacheTestSuite) TestReadFileWithRefresh() {
 	defer suite.cleanupTest()
-	// Configure to create empty files so we create the file in storage
+	// Configure to create empty files so we create the file in cloud storagetorage
```
Dabnsky, line 1412: typo

component/libfuse/libfuse2_handler.go
```diff
@@ -845,20 +835,20 @@ func (cf *CgofuseFS) Rename(oldpath string, newpath string) int {
...
-	if (dstErr == nil || os.IsExist(dstErr)) && !dstAttr.IsDir() && srcAttr.IsDir() {
+	if dstErr == nil && !dstAttr.IsDir() && srcAttr.IsDir() {
 		log.Err("Libfuse::Rename : dst [%s] is an existing file but src [%s] is a directory", dstPath, srcPath)
 		return -fuse.ENOTDIR
 	}
 
 	if srcAttr.IsDir() {
 		// ENOTEMPTY
-		if dstErr == nil || os.IsExist(dstErr) {
+		if dstErr == nil {
```
Dabnsky, line 851: why not use the && that the previous conditions are using?

component/s3storage/config.go
```diff
@@ -82,7 +82,9 @@ func ParseAndValidateConfig(s3 *S3Storage, opt Options) error {
 
 	// Part size must be at least 5 MB and smaller than 5GB. Otherwise, set to default.
 	if opt.PartSizeMb < 5 || opt.PartSizeMb > MaxPartSizeMb {
-		log.Warn("ParseAndValidateConfig : Part size must be between 5MB and 5GB. Default to default size")
+		if opt.PartSizeMb != 0 {
```
Dabnsky, line 85: why is opt.PartSizeMb being 0 important to check?

Dabnsky (approved):

component/file_cache/lru_policy.go
```diff
@@ -446,7 +445,13 @@ func (p *lruPolicy) deleteItem(name string) {
 	}
 
 	// There are no open handles for this file so its safe to remove this
-	err := deleteFile(name)
+	// Check if the file exists first, since this is often the second time we're calling deleteFile
+	_, err := os.Lstat(name)
+	if err == nil && os.IsNotExist(err) {
```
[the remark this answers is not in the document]
foodprocessor: Yes! `!=` was intended here. Good catch!
Regarding `Lstat`, I think you're right. Cargo cult detected! The original code (ab?)uses `Lstat`. I think we can safely change all those instances to `Stat`.

component/s3storage/config.go
foodprocessor: This code is meant to log a warning when the user sets `PartSizeMb` outside of the range we allow.
But when `PartSizeMb` is zero, that is supposed to mean that the user didn't set it at all, and they expected us to set it to default. In that case, nothing unexpected is happening, so I figured we shouldn't log a warning.
There is a slight chance that the user explicitly sets `PartSizeMb` to zero, but I'm fine with not logging a warning in that case, since setting a numeric value which needs to be nonzero to zero is a kind of standard way of saying "give me the default value."

component/attr_cache/attr_cache.go
foodprocessor: Good point!
This now has a brand new unit test.

jfantinhardesty (approved):

Dabnsky (approved):

PR #159, opened 2024-03-21
Add instructions for systemd

### LINKED ISSUE #80: Persistent mount on Linux
### Please describe the feature/improvement you would like to see added.

Mounted buckets are remounted at login on Windows. It would be nice to have the same functionality on Linux.

### Please explain why you think adding this feature/improvement is important.

Linux users would be more likely to be able to set up a process to remount their containers on their own, but that's not a guarantee. I see no reason why we should be less friendly to Linux users.

jfantinhardesty:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This adds correct instructions and additional documentation to have cloudfuse run using systemd.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #80


setup/readme.md
Dabnsky, line 1: We need an end-user to follow these instructions to give us feedback. any ideas on where to fine volunteers?

Dabnsky (approved):

setup/cloudfuse.service
```diff
@@ -0,0 +1,22 @@
...
+User=CloudfuseUser
+# Path to the location Cloudfuse will mount to. Note this folder must currently exist.
+Environment=MoutingPoint=/path/to/mounting/point
+# Path to the configuration file.
+Environment=ConfigFile=/path/to/config/file/config.yaml
+
+# Under the hood
+Type=forking
+ExecStart=/usr/bin/cloudfuse mount ${MoutingPoint} --config-file=${ConfigFile}
+ExecStop=/usr/bin/fusermount -u ${MoutingPoint}
```
foodprocessor, line 19: Should we make this a lazy unmount?

setup/readme.md
```diff
@@ -1,11 +1,14 @@
...
-2. Prepare the config file. Name of the container should be specified by `azstorage:container` in config file or AZURE_STORAGE_ACCOUNT_CONTAINER environment variable. 
-3. Download cloudfuse.service and put it into /etc/systemd/system
-4. Edit the file, changing environment values in Service block.
-5. Run command to reload service config files: `systemctl daemon-reload`
-6. Start service: `systemctl start cloudfuse.service`
-7. Make the service starting with system: `systemctl enable cloudfuse.service`
-8. Please Note that the example has the User AzureUser, please create a user called AzureUser or replace this with an existing user.
+## Steps to install Cloudfuse to systemd
+
+1. Prepare the configuration file `config.yaml` that cloudfuse will use to start your mount. Follow the general instructions for creating a config.yaml file from the Readme.
```
foodprocessor, line 5: Can you link back to the config instructions here?

setup/readme.md
```diff
@@ -1,11 +1,14 @@
...
+2. In the cloudfuse.service file, edit all of the fields under the `Service` section and replace them for your system. Please note that the example has the User CloudfuseUser, please create a user called CloudfuseUser or replace this with an existing user.
+3. Copy the cloudfuse.service and place it in /etc/systemd/system:
+   `cp cloudfuse.service /etc/systemd/system`
```
foodprocessor, line 8: I think this, and all the `systemctl` commands should have `sudo`

setup/readme.md
```diff
@@ -1,11 +1,14 @@
...
-3. Download cloudfuse.service and put it into /etc/systemd/system
-4. Edit the file, changing environment values in Service block.
-5. Run command to reload service config files: `systemctl daemon-reload`
-6. Start service: `systemctl start cloudfuse.service`
-7. Make the service starting with system: `systemctl enable cloudfuse.service`
-8. Please Note that the example has the User AzureUser, please create a user called AzureUser or replace this with an existing user.
+## Steps to install Cloudfuse to systemd
+
+1. Prepare the configuration file `config.yaml` that cloudfuse will use to start your mount. Follow the general instructions for creating a config.yaml file from the Readme.
+2. In the cloudfuse.service file, edit all of the fields under the `Service` section and replace them for your system. Please note that the example has the User CloudfuseUser, please create a user called CloudfuseUser or replace this with an existing user.
```
foodprocessor, line 6: I think we need to add the cloudfuse.service file to our deployment pipeline so it makes it into our Linux builds.
Relatedly, I'm not sure: should we push it to /etc/systemd ourselves? That might be the common practice.

foodprocessor (commented):
Looks good, but I got a SegFault!

foodprocessor (approved):
J/k it works now!

setup/readme.md
foodprocessor: I think cp needs it too.

setup/cloudfuse.service
jfantinhardesty: Yes, and this is fixed.

setup/readme.md
jfantinhardesty: Fixed.

setup/readme.md
jfantinhardesty: Fixed.

setup/readme.md
jfantinhardesty: I agree. After doing some reading it looks like we could write to /usr/lib/systemd/ which seems to be where these files should be placed for installed software https://unix.stackexchange.com/questions/206315/whats-the-difference-between-usr-lib-systemd-system-and-etc-systemd-system. I will add this to the backlog.

PR #160, opened 2024-03-22
Fix formatting testify tests

jfantinhardesty:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Fixes the formatting of our testify tests by using best practices from the testify linter https://github.com/Antonboom/testifylint. Additionally, this fixes the order of the assert expected and actual values so they are now all in the correct order. These were all automatically fixed by the linter so the changes should be correct.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


common/version_test.go
```diff
@@ -49,39 +49,39 @@ func (vSuite *versionTestSuite) TestVersionEquality() {
 
 	v1, _ := ParseVersion("10.0.0")
 	v2, _ := ParseVersion("10.0.0")
-	assert.Equal(v1.compare(*v2), 0)
+	assert.Equal(0, v1.compare(*v2))
```
Dabnsky, line 52: Did the linter make these changes as well? Seems like a trivial change here.

Dabnsky (approved):
I think I like these assert functions. they seem more convenient and intuitive.

common/version_test.go
jfantinhardesty: Yes. These changes to Equal are the most important ones in the PR. The Equal function expects that the first values is the expected value and the second is the actual value. When the tests are passing it doesn't matter the order, but when I test fails then the order being wrong will mislead you. Because the error message you will get is backward and it will tell you the wrong expected value and the wrong actual value.

foodprocessor (approved):
A delightful and complete improvement!

PR #181, opened 2024-04-05
Bugfix: File cache rename directory fails to remove source directory

foodprocessor:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Change file cache `invalidateDirectory` to delete files in order (children first)
Also added error logging for `os.RemoveAll`

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -399,29 +399,47 @@ func (fc *FileCache) invalidateDirectory(name string) {
 		return
 	}
 	// TODO : wouldn't this cause a race condition? a thread might get the lock before we purge - and the file would be non-existent
+	var directoryStack []string
 	err = filepath.WalkDir(localPath, func(path string, d fs.DirEntry, err error) error {
 		if err == nil && d != nil {
-			log.Debug("FileCache::invalidateDirectory : %s (IsDir=%t) getting removed from cache", path, d.IsDir())
+			// WalkDir goes through the tree in lexical order
+			// 	so 'aDir' always comes before 'aDir/file'
+			// 	and 'bDir' always comes after the whole 'aDir' subtree
+			// We can leverage this lexical order to queue folders for deletion after their children
+
+			// queue all files for deletion
 			if !d.IsDir() {
+				log.Debug("FileCache::invalidateDirectory : %s (file) getting removed from cache", path)
 				fc.policy.CachePurge(path)
-			} else {
-				deleteDirError := deleteFile(path)
-				if deleteDirError != nil {
-					log.Warn("FileCache::invalidateDirectory : %s - failed to delete %s. Here's why: %v", name, path, deleteDirError)
+			}
+			// queue folders for deletion after their children
+			if len(directoryStack) > 0 {
+				lastDirIndex := len(directoryStack) - 1
+				if !strings.HasPrefix(path, directoryStack[lastDirIndex]) {
+					log.Debug("FileCache::invalidateDirectory : %s (directory) getting removed from cache", path)
+					fc.policy.CachePurge(directoryStack[lastDirIndex])
+					directoryStack = directoryStack[:lastDirIndex]
 				}
+			}
+			// remember folders we've come across, to queue up later
+			if d.IsDir() {
+				directoryStack = append(directoryStack, path)
 			}
 		}
 		return nil
 	})
+	// queue up the remaining folders for deletion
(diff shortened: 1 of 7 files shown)


component/file_cache/file_cache.go
```diff
@@ -395,30 +399,48 @@ func (fc *FileCache) invalidateDirectory(name string) {
...
+			}
 		}
 		return nil
 	})
+	// queue up the remaining folders for deletion
+	for len(directoryStack) > 0 {
+		lastIndex := len(directoryStack) - 1
+		path := directoryStack[lastIndex]
+		log.Debug("FileCache::invalidateDirectory : %s (directory) getting removed from cache", path)
+		fc.policy.CachePurge(path)
```
jfantinhardesty, line 436: Will CachePurge handle directories for both lfu and lru? I thought that it only handled files, which is why delete was called separately for directories?

component/file_cache/file_cache.go
```diff
@@ -395,30 +399,48 @@ func (fc *FileCache) invalidateDirectory(name string) {
...
 			if !d.IsDir() {
+				log.Debug("FileCache::invalidateDirectory : %s (file) getting removed from cache", path)
 				fc.policy.CachePurge(path)
-			} else {
-				deleteDirError := deleteFile(path)
-				if deleteDirError != nil {
-					log.Warn("FileCache::invalidateDirectory : %s - failed to delete %s. Here's why: %v", name, path, deleteDirError)
+			}
+			// queue folders for deletion after their children
+			if len(directoryStack) > 0 {
```
jfantinhardesty, line 416: Can you explain why you need to also do this here, when you also do a purge of everything at the end of the function?

component/file_cache/file_cache.go
foodprocessor: For both policies, `CachePurge`  removes the path from the policy data structure, and then pushes the path into a queue for deletion.
Since folders are not considered to be cached items, when `CachePurge` goes to remove them from the policy records, it does nothing. But it still goes on to put them in the deletion queue.
The key to my bug fix here is that the folders need to be deleted in a queue - after their children - and not immediately (as they were doing before). So I am using `CachePurge` to achieve that without doing a lot of extra work creating a separate function.
Both policies implement `CachePurge` to ultimately call `os.Remove` on whatever is passed in. So I think this should work either way. But I have not tested it with LFU.

I honestly can't tell why they had different behavior for files and directories. Maybe it's related to the fact that part of the `CachePurge` functionality doesn't apply to directories... But it's honestly confusing, and I don't want to dissect the mistaken thinking behind it.

jfantinhardesty: Yeah I like the change, and the logic behind it. Maybe you could add a test for the LRU and LFU to test using CachePurge to delete a directory, to check that it works correctly.

component/file_cache/file_cache.go
foodprocessor: Technically I don't have to do this here. But it felt more satisfying to delete folders as soon as they're empty, and not all at the end.
But the code is more repetitive this way. I'll give you that.

For extra clarity, with our standard test nested folder structure:

- a/
  - a/c1/
    - a/c1/gc1
  - a/c2

this is the difference between this deletion sequence:

1. a/c1/gc1
2. a/c1/
3. a/c2
4. a

and this one:

1. a/c1/gc1
2. a/c2
3. a/c1/
4. a/

jfantinhardesty: Since it makes the code more repetitive, maybe you could just do it at the end, since it doesn't seem to improve performance. Also, since that the function is called for every file or directory, it seems inefficient to iterate over the list for each directory.

component/file_cache/file_cache.go
foodprocessor: I have added tests for LRU and LFU.

component/file_cache/file_cache.go
foodprocessor: I have just simplified this algorithm quite a bit. Since WalkDir is called in series, I just used it to build a list of items, which I then delete in reverse order. No more repetitive or complex logic. Just a bit more memory usage.

jfantinhardesty (approved):

PR #188, opened 2024-04-09
Support case3 for StreamDir

jfantinhardesty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

We now support case 3 for StreamDir, when a file exists in local storage and in the cloud. We will serve the correct modify time and size of the file based on the local cache. This has an additional cost of about 4ms per call to StreamDir. So for 10,000 items in our test bucket, it takes an extra 40ms to support this, which is not great but not too bad either. Let me know if you think this should be behind a flag in the file cache.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -442,8 +442,32 @@ func (fc *FileCache) DeleteDir(options internal.DeleteDirOptions) error {
 
 // StreamDir : Add local files to the list retrieved from storage container
 func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.ObjAttr, string, error) {
+	// For stream directory, there are three different child path situations we have to potentially handle.
+	// 1. Path in storage but not in local cache
+	// 2. Path not in storage but in local cache (this could happen if we recently created the file [and are currently writing to it]) (also supports immutable containers)
+	// 3. Path in storage and in local cache (this could result in dirty properties on the service if we recently wrote to the file)
+
+	// To cover case 1, grab all entries from storage
 	attrs, token, err := fc.NextComponent().StreamDir(options)
 
+	for _, attr := range attrs {
+		entryPath := common.JoinUnixFilepath(attr.Path)
+
+		// Return from local cache only if file is not under download or deletion
+		// If file is under download then taking size or mod time from it will be incorrect.
+		if !attr.IsDir() && !fc.fileLocks.Locked(entryPath) {
+			entryCachePath := common.JoinUnixFilepath(fc.tmpPath, entryPath)
+			info, err := os.Stat(entryCachePath)
+
+			// Case 3 (file is in storage and in local cache) so update the relevant attributes
+			if err == nil {
+				attr.Mtime = info.ModTime()
+				attr.Size = info.Size()
+			}
+		}
+	}
+
+	// To cover case 2, grab entries from the local cache
 	if token == "" {
 		// This is the last set of objects retrieved from container so we need to add local files here
 		localPath := common.JoinUnixFilepath(fc.tmpPath, options.Name)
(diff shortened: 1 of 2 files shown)


component/file_cache/file_cache.go
```diff
@@ -442,8 +442,32 @@ func (fc *FileCache) DeleteDir(options internal.DeleteDirOptions) error {
...
 func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.ObjAttr, string, error) {
+	// For stream directory, there are three different child path situations we have to potentially handle.
+	// 1. Path in storage but not in local cache
+	// 2. Path not in storage but in local cache (this could happen if we recently created the file [and are currently writing to it]) (also supports immutable containers)
+	// 3. Path in storage and in local cache (this could result in dirty properties on the service if we recently wrote to the file)
+
+	// To cover case 1, grab all entries from storage
 	attrs, token, err := fc.NextComponent().StreamDir(options)
 
+	for _, attr := range attrs {
```
foodprocessor, line 453: I think we optimize this by using a double-indexed loop, which walks results from `os.ReadDir` and `attrs` in parallel?
`attrs` is already alphabetized (I think!), and the `os.ReadDir` results can be alphabetized quickly using `slices.SortFunc`.
If you implement this suggestion, then since we call `os.ReadDir` to merge case 2 entries with the last chunk, maybe we could use a single `os.ReadDir` call and use the results for either thing based on whether we're on the last chunk.

component/file_cache/file_cache.go
```diff
@@ -442,8 +442,32 @@ func (fc *FileCache) DeleteDir(options internal.DeleteDirOptions) error {
...
+		entryPath := common.JoinUnixFilepath(attr.Path)
```
foodprocessor, line 454: Do we need to normalize the path here? I think the name here should already be normalized.

component/file_cache/file_cache.go
```diff
@@ -453,22 +477,23 @@ func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.O
...
-					// If entry does not exists on storage then only return a local item here.
-					_, err := fc.NextComponent().GetAttr(internal.GetAttrOptions{Name: entryPath})
-					if err != nil && (err == syscall.ENOENT || os.IsNotExist(err)) {
-						log.Debug("FileCache::StreamDir : serving %s from local cache", entryPath)
-						attr := newObjAttr(entryPath, info)
-						attrs = append(attrs, attr)
+
+				if !entry.IsDir() && !fc.fileLocks.Locked(entryPath) {
+					entryCachePath := common.JoinUnixFilepath(fc.tmpPath, entryPath)
+					info, err := os.Stat(entryCachePath) // Grab local cache attributes
```
foodprocessor, line 483: It might be easier to get this stat information by calling `entry.Info()` instead of calling `os.Stat` explicitly.
Doing that may also have a performance benefit, since [according to the docs](https://pkg.go.dev/io/fs#DirEntry), the `os.ReadDir()` results might be reused to provide this data.

component/file_cache/file_cache_test.go
```diff
@@ -390,6 +389,97 @@ func (suite *fileCacheTestSuite) TestStreamDirCase2() {
...
+	suite.assert.EqualValues(file1, dir[0].Path)
+	suite.assert.EqualValues(1024, dir[0].Size)
+	suite.assert.EqualValues(file2, dir[1].Path)
+	suite.assert.EqualValues(1024, dir[1].Size)
+	suite.assert.EqualValues(file3, dir[2].Path)
+	suite.assert.EqualValues(1024, dir[2].Size)
+	suite.assert.EqualValues(subdir, dir[3].Path)
+}
+
+func pos(s []*internal.ObjAttr, e string) int {
```
foodprocessor, line 428: Very nice!

foodprocessor (changes requested):
I think with some optimization, maybe we could get the penalty down enough to avoid a flag!

component/file_cache/file_cache.go
```diff
@@ -453,22 +477,23 @@ func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.O
...
+					if err == nil {
+						// This is an overhead for streamdir for now
+						// As list is paginated we have no way to know whether this particular item exists both in local cache
+						// and container or not. So we rely on getAttr to tell if entry was cached then it exists in cloud storage too
+						// If entry does not exists on storage then only return a local item here.
+						_, err := fc.NextComponent().GetAttr(internal.GetAttrOptions{Name: entryPath})
+						if err != nil && (err == syscall.ENOENT || os.IsNotExist(err)) {
+							// Case 2 (file only in local cache) so create a new attributes and add them to the storage attributes
+							log.Debug("FileCache::StreamDir : serving %s from local cache", entryPath)
+							attr := newObjAttr(entryPath, info)
```
foodprocessor, line 494: Since `info` is only used when `GetAttr` returns an error, can we fetch `info` here instead of doing so before calling `GetAttr`?

component/file_cache/file_cache.go
jfantinhardesty: Yes, using os.ReadDir is faster than the stat. I used your suggestion of a double-indexed loop and it is faster on Linux compared to the stat method when the cache is small. However, once the cache is larger, then this method will be slower. It is about a .5 -1ms penalty for every 500 files in the cache. So as long as the number of items in a folder is less than ~2000, this is better. On Windows, performance is significantly improved, with about a 1-2ms penalty per 500 files.

component/file_cache/file_cache.go
jfantinhardesty: Good catch.

component/file_cache/file_cache.go
jfantinhardesty: Thanks for the suggestion. Yeah it seems .Info could be slightly faster, so I moved it all to .Info().

component/file_cache/file_cache.go
jfantinhardesty: Yes, good catch. This will be much faster.

foodprocessor (approved):
Great!
If we see performance complaints down the road, we can always flag this, but I think this is a very good solution to a complex problem.
Thank you!

Dabnsky (approved):

PR #197, opened 2024-04-20
Fix symlink support

foodprocessor:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Change `no-symlinks` config option to `enable-symlinks` and change default to disable symlinks.
Always pull symlink option from `attr_cache` into cloud storage component.
Clean up s3storage tests.
Return `syscall.ENOTSUP` on `CreateLink` and `ReadLink` when symlinks are disabled.
**`azstorage` was not changed (will not return `syscall.ENOTSUP` on link calls). This may need to change (please review).**

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/s3storage/s3storage_test.go
+++ b/component/s3storage/s3storage_test.go
@@ -2391,6 +2391,10 @@ func (s *s3StorageTestSuite) TestGetAttrFile() {
 
 func (s *s3StorageTestSuite) TestGetAttrLink() {
 	defer s.cleanupTest()
+	// enable symlinks in config
+	config := generateConfigYaml(storageTestConfigurationParameters) + "attr_cache:\n  enable-symlinks: true\n"
+	s.setupTestHelper(config, s.bucket, true)
+	s.assert.False(s.s3Storage.stConfig.disableSymlink)
 	// Setup
 	target := generateFileName()
 	s.s3Storage.CreateFile(internal.CreateFileOptions{Name: target})
--- a/component/s3storage/s3storage_test.go
+++ b/component/s3storage/s3storage_test.go
@@ -2301,7 +2301,7 @@ func (s *s3StorageTestSuite) TestReadLinkDisabled() {
 
 	_, err := s.s3Storage.ReadLink(internal.ReadLinkOptions{Name: name})
 	s.assert.Error(err)
-	s.assert.EqualValues(syscall.ENOTSUP, err)
+	s.assert.EqualValues(syscall.ENOENT, err)
 }
 
 func (s *s3StorageTestSuite) TestGetAttrDir() {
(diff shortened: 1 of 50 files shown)


jfantinhardesty (commented):
Some of the tests for S3 storage are broken, so I think there may be some more tests that rely on symlink to be true. Is there a particular reason you did not change azstorage to not return syscall.ENOTSUP for CreateLink and ReadLink. It seems that it should be consistent with s3storage?

component/s3storage/s3storage_test.go
```diff
@@ -2263,44 +2294,37 @@ func (s *s3StorageTestSuite) TestReadLinkError() {
...
+	name := generateFileName()
+
+	_, err := s.s3Storage.ReadLink(internal.ReadLinkOptions{Name: name})
+	s.assert.Error(err)
+	s.assert.EqualValues(syscall.ENOTSUP, err)
+}
+
 func (s *s3StorageTestSuite) TestGetAttrDir() {
 	defer s.cleanupTest()
-	vdConfig := generateConfigYaml(storageTestConfigurationParameters)
```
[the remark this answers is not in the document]
foodprocessor: When I was figuring out how to change the config to enable symlinks in the tests that need them, I noticed that `setupTestHelper` sets some config parameters by default when an empty config string is passed in. To do that, it uses `generateConfigYaml(storageTestConfigurationParameters)`... but that is exactly the same as `vdConfig`.
I realized the reason for this is that we copied these tests from `azstorage`, where `vdConfig` was used to set `virtual-directory` to `true`.
But `s3storage` doesn't have a `virtual-directory` setting, so that's why these `vdConfig` variables are just redundantly testing the same default settings a second time.
Once I figured all that out, I decided to tear out all the virtual directory stuff from the tests in `s3storage`.

foodprocessor:
> Some of the tests for S3 storage are broken, so I think there may be some more tests that rely on symlink to be true. Is there a particular reason you did not change azstorage to not return syscall.ENOTSUP for CreateLink and ReadLink. It seems that it should be consistent with s3storage?

I think I had some hesitation to change `azstorage` for some reason, even though it needed to be changed.
I have now added identical error returns to the link functions in `azstorage`.

doc/cloudfuse_mount.md
```diff
@@ -47,7 +47,7 @@ cloudfuse mount [path] [flags]
       --negative-timeout uint32           The negative entry timeout in seconds.
       --network-share                     Run as a network share. Only supported on Windows.
       --no-cache-dirs                     whether or not empty directories should be cached
-      --no-symlinks                       whether or not symlinks should be supported
+      --enable-symlinks                       whether or not symlinks should be supported
```
jfantinhardesty, line 50: Could you fix the spacing here.

jfantinhardesty (approved):
Looks good!

doc/cloudfuse_mount.md
foodprocessor: Woops! Fixed.

PR #212, opened 2024-05-17
Fix AWS Region and Endpoint precendence

jfantinhardesty:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Currently, using the AWS_REGION environment variable was not supported and would not work correctly. This refactors the endpoint resolution and allows the AWS_REGION flag to be used. Additionally, this adds endpoint, region, and bucket-name to be passed in as command line parameters.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


doc/cloudfuse_mount_list.md
```diff
@@ -37,31 +37,34 @@ cloudfuse mount list
...
+      --endpoint string                   S3 endpoint url.
       --entry-timeout uint32              The entry timeout in seconds.
       --file-cache-timeout uint32         file cache timeout (default 120)
       --foreground                        Mount the system in foreground mode. Default value false.
       --hard-limit                        File cache limits are hard limits or not.
       --high-disk-threshold uint32        percentage of cache utilization which kicks in early eviction (default 90)
       --ignore-open-flags                 Ignore unsupported open flags (APPEND, WRONLY) by cloudfuse when writeback caching is enabled. (default true)
       --ignore-sync                       Just ignore sync call and do not invalidate locally cached file.
-      --log-file-path string              Configures the path for log files. Default is $HOME/.cloudfuse/cloudfuse.log (default "$HOME/.cloudfuse/cloudfuse.log")
+      --log-file-path string              Configures the path for log files. Default is /home/jfan/.cloudfuse/cloudfuse.log (default "/home/jfan/.cloudfuse/cloudfuse.log")
```
foodprocessor, line 57: This looks like a bug in the documentation generation

component/s3storage/client.go
```diff
@@ -138,15 +106,18 @@ func (cl *Client) Configure(cfg Config) error {
...
+		})
 	}
 
-	// ListBuckets here to test connection
-	_, err = cl.ListBuckets()
+	// List objects to test connection
+	_, _, err = cl.List("/", nil, 1)
 	if err != nil {
-		log.Err("Client::Configure : listing buckets failed. Here's why: %v", err)
+		log.Err("Client::Configure : listing objects failed. Here's why: %v", err)
```
foodprocessor, line 120: I'm a bit confused about this... Where is the bucket set?
And is there a reason this needs to change?
If we default to the first visible bucket, we should probably only do that when the bucket-name config option is empty.

component/s3storage/client.go
```diff
@@ -121,14 +77,26 @@ func (cl *Client) Configure(cfg Config) error {
 			"",
 		)
 	}
+
+	if cl.Config.authConfig.Region == "" {
+		_, exists := os.LookupEnv("AWS_REGION")
+		if !exists {
+			cl.Config.authConfig.Region = "us-east-1"
+		}
+	}
```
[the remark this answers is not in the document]
jfantinhardesty: This is fixed. I added code to get the region from the endpoint, if the region was not provided.

doc/cloudfuse_mount_list.md
jfantinhardesty: Fixed.

component/s3storage/client.go
jfantinhardesty: The bucket is set whenever we make an S3 call and it is set in the config. With the new default to the first bucket, this additional list call validates that the region is correct (which ListBuckets does not do) and validates that we have access to the bucket.

component/s3storage/client.go
```diff
@@ -155,7 +148,30 @@ func (cl *Client) Configure(cfg Config) error {
...
+
+	return nil
+}
+
+func getRegionFromEndpoint(endpoint string) (string, error) {
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return "", err
+	}
```
foodprocessor, line 165: We should explicitly check for an empty endpoint, since that won't return an error here.

foodprocessor (changes requested):
I've just reviewed baseConfig.yaml & client.go for now.
More to come!

foodprocessor:
> I've just reviewed baseConfig.yaml & client.go for now. More to come!

Just kidding, the rest looks good!

Dabnsky (approved):

setup/baseConfig.yaml
```diff
@@ -179,7 +179,7 @@ s3storage:
   bucket-name: <name of the bucket to be mounted. Default - first accessible bucket>
   key-id: <S3 access key ID. Default - use credentials from environment variables or shared profile (see README)>
   secret-key: <S3 secret access key. Default - use credentials from environment variables or shared profile (see README)>
-  region: <S3 region. Default - us-east-1>
+  region: <S3 region. If endpoint is provided, will read region from endpoint. Default - us-east-1>
```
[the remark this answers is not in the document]
jfantinhardesty: I tried to rephrase this a bit better. The basic idea is that if the user does not specify a region, then if the user provides an endpoint it uses the region in that. Otherwise, it uses the default.

component/s3storage/client.go
```diff
@@ -155,7 +148,30 @@ func (cl *Client) Configure(cfg Config) error {
...
+
+	hostParts := strings.Split(u.Hostname(), ".")
+	// Region should be the after the first period
+	// Ex: https://s3.us-east-1.lyvecloud.seagate.com
+	if len(hostParts) < 2 {
+		return "", fmt.Errorf("Unable to parse Endpoint region")
+	}
+	region := hostParts[1]
```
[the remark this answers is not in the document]
jfantinhardesty: Thanks for the suggestion. Changed and added several of these to the tests.

component/s3storage/client.go
jfantinhardesty: Good point.

setup/baseConfig.yaml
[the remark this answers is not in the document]
foodprocessor: I think this is clearer. Thanks!

foodprocessor (approved):
Thank you!

PR #219, opened 2024-05-23
Delete faster file cache

Dabnsky:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

These changes optimizes the deletions and other open() calls that the windows OS invokes on files that don't require downloaded data.

summarized changes: 

Moved the download logic out of the `filecache.OpenFile()` into `downloadFile()`. `OpenFile()` now spits out an empty handle with flags and filemode data stored in the handle.values map that `downloadFile()` uses.

`downloadFile()` is invoked by `ReadInBuffer()` and `WriteFile()` when the handle.value exists.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


component/file_cache/file_cache.go
```diff
@@ -691,9 +692,47 @@ func (fc *FileCache) DeleteFile(options internal.DeleteFileOptions) error {
...
+
+	handle.RemoveValue("flag")
+	handle.RemoveValue("mode")
+
+	fc.mu.Unlock()
+
+	return handle
+}
+
+func (fc *FileCache) DownloadFile(options internal.DownloadFileOptions) (*handlemap.Handle, error) {
```
foodprocessor, line 734: You could pass the whole handle into this function, and then you could modify its fields instead of replacing the whole struct.

component/file_cache/file_cache.go
```diff
@@ -704,107 +743,172 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
...
+	f, err = common.OpenFile(localPath, options.Flags, options.Mode)
+	if err != nil {
+		log.Err("FileCache::download : error opening cached file %s [%s]", options.Name, err.Error())
+		return nil, err
+	}
+
+	// Increment the handle count in this lock item as there is one handle open for this now
+	flock.Inc()
+
+	handle := handlemap.NewHandle(options.Name)
```
foodprocessor, line 867: If you pass the handle into `DownloadFile`, you could update the struct fields by simply skipping this line.
Then the lines below would update the existing handle instead of creating a new one.

foodprocessor (changes requested):
This looks good!
I think we're ready to move on to dealing with concurrency!
So, my comments are all about that.

component/file_cache/file_cache.go
```diff
@@ -1268,11 +1380,28 @@ func (fc *FileCache) TruncateFile(options internal.TruncateFileOptions) error {
...
+				return err
+			}
+		} else {
+			h, err = fc.OpenFile(internal.OpenFileOptions{Name: options.Name, Flags: os.O_RDWR, Mode: fc.defaultPermission})
+			if err != nil {
+				log.Err("FileCache::TruncateFile : Error opening file %s [%s]", options.Name, err.Error())
+				return err
+			}
+		}
+
```
foodprocessor, line 1404: Shouldn't this code just call `fc.OpenFile`, followed by the same code you use in `ReadInBuffer` and `Write` to call `getHandleData`?

component/file_cache/file_cache.go
```diff
@@ -704,107 +739,171 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
 	defer flock.Unlock()
```
foodprocessor, line 739: This flock is per-file (not just per-handle), so once it locks, we know that we're the only ones working on this file.
This could be a good opportunity to address the potential issue of two handles trying to download the same file.
Maybe you could call `isDownloadRequired` (again) here, and if it comes back false, that means another handle has already downloaded the file. In that case, you could skip downloading it and just open the already-downloaded file.

foodprocessor (changes requested):
The updated handle code looks good!
Now I've added comments about per-file concurrency to prevent two different handles from downloading the same file.

component/file_cache/file_cache.go
```diff
@@ -704,107 +739,171 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
...
+
+	attr, err := fc.NextComponent().GetAttr(internal.GetAttrOptions{Name: options.Name})
+	if err != nil {
+		log.Err("FileCache::download : Failed to get attr of %s [%s]", options.Name, err.Error())
+	}
 
 	// return err in case of authorization permission mismatch
 	if err != nil && err == syscall.EACCES {
 		return nil, err
 	}
```
foodprocessor, line 751: Since our only opportunity to return errors like `EACCES` is in response to `OpenFile`, this code should probably move there.
Any other code that returns an error should move there too, if it can (if it doesn't require the file to be downloaded first).

foodprocessor (changes requested):
This one is a suggestion to reorganize your function calls to improve readability.

component/file_cache/file_cache.go
foodprocessor: Okay, I guess this comment isn't about concurrency. But the next one is!

component/file_cache/file_cache.go
```diff
@@ -691,9 +691,44 @@ func (fc *FileCache) DeleteFile(options internal.DeleteFileOptions) error {
...
+
+	// extract the filemode out of handle values
+	mode, _ := handle.GetValue("mode")
+	fileModeValue, ok := mode.(os.FileMode)
+	if !ok {
+		log.Debug("FileCache::getHandleData : error Type assertion failed on getting file mode for %s [%s]", handle.Path)
+		return nil
+	}
+
+	handle, err := fc.DownloadFile(internal.DownloadFileOptions{Name: handle.Path, Flags: flag, Mode: fileModeValue, Handle: handle})
```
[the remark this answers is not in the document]
Dabnsky: getHandleData is more about fetching the handle data that's initially missing. It goes and downloads and loads in the rest of the handle data that wasn't there previously. The flag and mode is a kind of assurance to make sure the correct permissions and filemode is used while the file being used for the download has the correct configuration.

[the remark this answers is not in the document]
foodprocessor: Ok, now I get why it's called that.
I do feel like `getHandleData` is a bit vague, compared to `DonwloadFile`, which is direct and clear.
But anyway, this comment is just about simplifying and improving readability, which is less important than the concurrency and error handling issues above.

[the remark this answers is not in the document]
Dabnsky: I'm open to better names if anyone has a suggestion. I'll give it some thought. downloadHandleData seems a bit long.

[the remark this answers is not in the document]
Dabnsky: Since the getHandleData is so small, It would be easy to incorperate the same code into DownloadFile() and get rid of getHandleData. I'll go ahead and apply what you're suggesting. Just like bad jokes, if it needs explaining, then it needs work.

component/file_cache/file_cache.go
```diff
@@ -691,69 +691,99 @@ func (fc *FileCache) DeleteFile(options internal.DeleteFileOptions) error {
...
+	//extract flag out of the values from handle
+	flags, found := options.Handle.GetValue("flag")
+	if found {
+		flagsStruct, ok := flags.(struct{ flags int })
+		if !ok {
+			log.Err("FileCache::getHandleData : error Type assertion failed on getting flag for %s", options.Handle.Path)
+			return options.Handle, fmt.Errorf("type assertion failed on getting flag forfor %s", options.Handle.Path)
+		}
+		flag = flagsStruct.flags
+	}
```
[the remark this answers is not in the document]
Dabnsky: The lack of flag means that it is cached. The flag in place means it needs to be downloaded

component/file_cache/file_cache.go
```diff
@@ -780,31 +810,90 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
...
+	options.Handle.UnixFD = uint64(f.Fd())
+	if !fc.offloadIO {
+		options.Handle.Flags.Set(handlemap.HandleFlagCached)
+	}
+
+	log.Info("FileCache::download : file=%s, fd=%d", options.Name, f.Fd())
+	options.Handle.SetFileObject(f)
+
+	//remove values as a kind of signal that the file has been downloaded
+	options.Handle.RemoveValue("flag")
```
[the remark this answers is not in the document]
Dabnsky: I could add another value that's simply another downloadRequired bool and check that instead. Does that sound good?

component/file_cache/file_cache.go
```diff
@@ -1268,11 +1374,25 @@ func (fc *FileCache) TruncateFile(options internal.TruncateFileOptions) error {
...
+			log.Err("FileCache::TruncateFile : Error calling OpenFile with %s [%s]", options.Name, err.Error())
 		}
+		if _, loaded := h.GetValue("flag"); loaded {
+			_, err = fc.DownloadFile(internal.DownloadFileOptions{Name: options.Name, Handle: h})
+			if err != nil {
+				log.Err("FileCache::TruncateFile : Error opening file %s [%s]", options.Name, err.Error())
+				return err
+			}
+		} else {
+			h, err = fc.OpenFile(internal.OpenFileOptions{Name: options.Name, Flags: os.O_RDWR, Mode: fc.defaultPermission})
```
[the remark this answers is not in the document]
Dabnsky: when the file gets flushed, it needs write permissions. I just updated the initial OpenFile() call to also include RDWR flag before it gets downloaded. the flags are now consistent. Does this seem correct to you?

component/file_cache/file_cache.go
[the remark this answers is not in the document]
jfantinhardesty: Yes, but if it does not have a flag, then it seems like you would not need to continue with the rest of the function since you would know here the file is cached. Also, I am not sure how you use the flag variable here. You set flag := 0 and in the case it is not found, then it is set to 0 and later passed into an OpenFile call, which seems incorrect in the case where the file is already cached.

component/file_cache/file_cache.go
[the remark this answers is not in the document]
jfantinhardesty: Yeah, that might be a little cleaner to read.

component/file_cache/file_cache.go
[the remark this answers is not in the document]
jfantinhardesty: Yes, but I don't see why you need an open call on 1378 and then another open call on line 1389.

component/file_cache/file_cache.go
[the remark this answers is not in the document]
Dabnsky: I was letting it download anyway due to some tests that were now calling DownloadFile() directly instead of OpenFile() as a way to download. the TestHardLimitOnSize was failing because originally, it was returning a nil handle once it failed to find the values in the handle.

I could have called OpenFile() before calling DownloadFile() in the test to potentially remedy this. Would that be a more preferable approach?

For the flag, what is a better default value to provide? I think I actually need to use the os.O_RD instead of a hard number like I was doing.

component/file_cache/file_cache.go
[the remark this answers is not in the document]
Dabnsky: the first `OpenFile()` call is for the `DownloadFile()` to take place. This came from Michael's suggestion to call Open instead of creating a new Handle directly like I was doing previously. 

The second `OpenFile()` call might not be necessary since it was already called previously. That was there previously. The entire else block might not be needed anymore. Do you agree?

component/file_cache/file_cache.go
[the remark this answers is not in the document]
jfantinhardesty: Yes, I think having the test do OpenFile and then Download is preferable. We shouldn't change out code to make the tests pass. And every time we interact with a file OpenFile is always called first, so that is a reasonable change to this test, and what it was doing before since OpenFile did previously download the file. And I don't think you need a default for the flag here, if you find the flag, then you use that value. If you don't then it didn't need to be downloaded anyway.

component/file_cache/file_cache.go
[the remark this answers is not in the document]
jfantinhardesty: After looking at this again, I think you can simplify it further. Here I think you are trying to think about it too hard. So you are determining here is the file needs to be downloaded, but there is a ton of logic built into DownloadFile already to figure that out for you. So don't worry about it here. Just call the first OpenFile and then DownloadFile after. It will download it if needed, and it will skip it if not.

component/file_cache/file_cache.go
```diff
@@ -920,6 +1022,24 @@ func (fc *FileCache) ReadInBuffer(options internal.ReadInBufferOptions) (int, er
 	// The file should already be in the cache since CreateFile/OpenFile was called before and a shared lock was acquired.
 	// log.Debug("FileCache::ReadInBuffer : Reading %v bytes from %s", len(options.Data), options.Handle.Path)
 
+	options.Handle.Lock()
+	var err error
+	if dwnldIntfce, found := options.Handle.GetValue("isDownloadNeeded"); found {
```
jfantinhardesty, line 1027: I think these other checks for Download could be simplified since I believe this logic is already in DownloadFile.

component/file_cache/file_cache.go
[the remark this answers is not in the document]
jfantinhardesty: I might suggest running this by Michael just to flesh out the design a bit, but my thinking is that you should either have DownloadFile be responsible for checking if a download is required or not (probably prefereable) or the caller of DownloadFile is responsible for checking if a download is required and then it will download it. Right now you are doing a bit of both.

component/file_cache/file_cache.go
```diff
@@ -920,6 +961,14 @@ func (fc *FileCache) ReadInBuffer(options internal.ReadInBufferOptions) (int, er
 	// The file should already be in the cache since CreateFile/OpenFile was called before and a shared lock was acquired.
 	// log.Debug("FileCache::ReadInBuffer : Reading %v bytes from %s", len(options.Data), options.Handle.Path)
 
+	options.Handle.Lock()
+	var err error
+	options.Handle, err = fc.DownloadFile(internal.DownloadFileOptions{Name: options.Handle.Path, Handle: options.Handle})
```
foodprocessor, line 966: Now that all you need to pass in to `DownloadFile` is the handle, maybe you can drop `DownloadFileOptions` and just pass the handle on its own?

component/libfuse/libfuse2_handler.go
```diff
@@ -642,6 +642,7 @@ func (cf *CgofuseFS) Open(path string, flags int) (int, uint64) {
 
 // Read reads data from a file into the buffer with the given offset.
 func (cf *CgofuseFS) Read(path string, buff []byte, ofst int64, fh uint64) int {
+	log.Debug("Libfuse::Read : reading path %s, handle: %d", path, fh)
```
foodprocessor, line 645: Log output was intentionally skipped in `Read` and `Write` to avoid creating log noise and the performance costs of a huge number of calls.
You could just drop this log call, or maybe the most readable thing would be to comment it out and add a comment explaining why we don't have log output here.

component/libfuse/libfuse2_handler.go
```diff
@@ -680,6 +681,7 @@ func (cf *CgofuseFS) Read(path string, buff []byte, ofst int64, fh uint64) int {
 
 // Write writes data to a file from the buffer with the given offset.
 func (cf *CgofuseFS) Write(path string, buff []byte, ofst int64, fh uint64) int {
+	log.Debug("Libfuse::Write : Writing path %s, handle: %d", path, fh)
```
foodprocessor, line 684: Same as above

internal/component_options.go
```diff
@@ -77,6 +77,11 @@ type DeleteFileOptions struct {
 	Name string
 }
 
+type DownloadFileOptions struct {
+	Name   string
+	Handle *handlemap.Handle
+}
+
```
foodprocessor, line 84: Since the name is already in `handle.Path`, you could drop this struct altogether.
This comment is the other half of the comment I left on `DownloadFile` above.

component/file_cache/file_cache.go
```diff
@@ -780,61 +799,82 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
...
+	log.Trace("FileCache::OpenFile : name=%s, flags=%d, mode=%s", options.Name, options.Flags, options.Mode)
+
+	_, err := fc.NextComponent().GetAttr(internal.GetAttrOptions{Name: options.Name})
+
+	// return err in case of authorization permission mismatch
+	if err != nil && err == syscall.EACCES {
+		return nil, err
+	}
+
+	fileCacheStatsCollector.UpdateStats(stats_manager.Increment, dlFiles, (int64)(1))
```
foodprocessor, line 869: The original code has two different versions of this call, with `dlFiles` when it downloads the file, and `cacheServed` when it doesn't.
So this call probably needs to move to `DownloadFile`.

component/file_cache/file_cache.go
```diff
@@ -691,69 +691,88 @@ func (fc *FileCache) DeleteFile(options internal.DeleteFileOptions) error {
 	return nil
 }
 
-// OpenFile: Makes the file available in the local cache for further file operations.
-func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle, error) {
-	log.Trace("FileCache::OpenFile : name=%s, flags=%d, mode=%s", options.Name, options.Flags, options.Mode)
+func (fc *FileCache) DownloadFile(options internal.DownloadFileOptions) (*handlemap.Handle, error) {
+
+	//extract flag and mode out of the value from handle
+	var flag int
```
foodprocessor, line 697: `flag` should probably always be plural, since the integer holds one flag in each bit.

component/file_cache/file_cache.go
```diff
@@ -691,69 +691,88 @@ func (fc *FileCache) DeleteFile(options internal.DeleteFileOptions) error {
...
+
+	flagMode, found := options.Handle.GetValue("fileFlagMode")
+	if found {
+		flagModeStruct, ok := flagMode.(struct {
+			flag  int
+			fMode fs.FileMode
+		})
+		if !ok {
+			log.Err("FileCache::DownloadFile : error Type assertion failed on getting flag for %s", options.Handle.Path)
+			return options.Handle, fmt.Errorf("type assertion failed on getting flag forfor %s", options.Handle.Path)
```
foodprocessor, line 708: typo: `forfor`

component/file_cache/file_cache.go
```diff
@@ -691,69 +691,88 @@ func (fc *FileCache) DeleteFile(options internal.DeleteFileOptions) error {
...
-	log.Trace("FileCache::OpenFile : name=%s, flags=%d, mode=%s", options.Name, options.Flags, options.Mode)
+func (fc *FileCache) DownloadFile(options internal.DownloadFileOptions) (*handlemap.Handle, error) {
+
+	//extract flag and mode out of the value from handle
+	var flag int
+	var fMode fs.FileMode
+
+	flagMode, found := options.Handle.GetValue("fileFlagMode")
+	if found {
+		flagModeStruct, ok := flagMode.(struct {
```
foodprocessor, line 702: `flags` and `mode` are the information needed to open the file once it's downloaded.
If the name of the field, struct, and variable they're in reflected their purpose, it'd probably be clearer. Maybe `openFileOptions`?

component/file_cache/file_cache.go
```diff
@@ -691,69 +691,88 @@ func (fc *FileCache) DeleteFile(options internal.DeleteFileOptions) error {
...
-// OpenFile: Makes the file available in the local cache for further file operations.
-func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle, error) {
-	log.Trace("FileCache::OpenFile : name=%s, flags=%d, mode=%s", options.Name, options.Flags, options.Mode)
+func (fc *FileCache) DownloadFile(options internal.DownloadFileOptions) (*handlemap.Handle, error) {
+
+	//extract flag and mode out of the value from handle
+	var flag int
+	var fMode fs.FileMode
+
+	flagMode, found := options.Handle.GetValue("fileFlagMode")
```
foodprocessor, line 700: `flagMode` and `flagModeStruct` have confusingly similar names. Since the first one is really just a temporary variable for untyped data, you could use the pattern we have in other places where `GetValue` is called, where the first untyped throw-away variable is just called `val`, and then the meaningful variable name is saved for when the type is asserted.

foodprocessor (changes requested):
Now just the little stuff!

component/file_cache/file_cache.go
```diff
@@ -1268,11 +1325,18 @@ func (fc *FileCache) TruncateFile(options internal.TruncateFileOptions) error {
 	} else {
 		// If size is not 0 then we need to open the file and then truncate it
 		// Open will force download if file was not present in local system
+
 		h, err = fc.OpenFile(internal.OpenFileOptions{Name: options.Name, Flags: os.O_RDWR, Mode: fc.defaultPermission})
+		if err != nil {
+			log.Err("FileCache::TruncateFile : Error calling OpenFile with %s [%s]", options.Name, err.Error())
+		}
+
+		h, err = fc.DownloadFile(internal.DownloadFileOptions{Name: options.Name, Handle: h})
```
foodprocessor, line 1334: Since the handle doesn't change anymore, `DownloadFile` doesn't need to return it.

Dabnsky: `DownlaodFile `will be receiving an empty handle here (that came from `OpenFile `and needs to return the handle with updated information.

foodprocessor: The handle is a pointer, so `h` in the current scope is already updated when `DownloadFile` writes to it.

Dabnsky: Thanks. Getting used to pointers

component/file_cache/file_cache.go
```diff
@@ -691,69 +696,85 @@ func (fc *FileCache) DeleteFile(options internal.DeleteFileOptions) error {
...
+	var flags int
+	var fMode fs.FileMode
+
+	flagMode, found := handle.GetValue("fileFlagMode")
+	if found {
+		fileOptions, ok := flagMode.(openFileOptions)
+		if !ok {
+			log.Err("FileCache::DownloadFile : error Type assertion failed on getting flag for %s", handle.Path)
+			return handle, fmt.Errorf("type assertion failed on getting flag for %s", handle.Path)
+		}
```
foodprocessor, line 711: I think this type assertion should follow the pattern set in other places where `GetValue` is used. Like in libfuse2_handler.go.

foodprocessor (commented):
Looks good!

foodprocessor (changes requested):
🎉

jfantinhardesty (commented):
I have been testing this branch on Windows and have noticed that many calls to close file fail with an error "FileCache::CloseFile : error [missing fd in handle object]." This seems to happen when there is an Open, Flush, Release call to libfuse. So there still might be a bit more to iron out in CloseFile.

component/file_cache/file_cache.go
```diff
@@ -1267,8 +1316,13 @@ func (fc *FileCache) TruncateFile(options internal.TruncateFileOptions) error {
...
-		// Open will force download if file was not present in local system
+		// downloadFile will provide download if file was not present in local system
 		h, err = fc.OpenFile(internal.OpenFileOptions{Name: options.Name, Flags: os.O_RDWR, Mode: fc.defaultPermission})
+		if err != nil {
+			log.Err("FileCache::TruncateFile : Error calling OpenFile with %s [%s]", options.Name, err.Error())
+		}
+
+		err = fc.downloadFile(h)
 		if err != nil {
 			log.Err("FileCache::TruncateFile : Error opening file %s [%s]", options.Name, err.Error())
```
foodprocessor, line 1327: This log message is perefect for the call above, but here it should probably mention `downloadFile`.

component/file_cache/file_cache.go
```diff
@@ -920,6 +955,14 @@ func (fc *FileCache) ReadInBuffer(options internal.ReadInBufferOptions) (int, er
 	// The file should already be in the cache since CreateFile/OpenFile was called before and a shared lock was acquired.
 	// log.Debug("FileCache::ReadInBuffer : Reading %v bytes from %s", len(options.Data), options.Handle.Path)
 
+	options.Handle.Lock()
+	var err error
+	err = fc.downloadFile(options.Handle)
+	if err != nil {
+		return 0, fmt.Errorf("error downloading file for %s [%s]", options.Handle.Path, err)
```
foodprocessor, line 962: The word "for" here seems extra.

foodprocessor (changes requested):
Very close!

component/file_cache/file_cache.go
```diff
@@ -831,9 +839,30 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
...
+// OpenFile: Makes the file available in the local cache for further file operations.
+func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle, error) {
+	log.Trace("FileCache::OpenFile : name=%s, flags=%d, mode=%s", options.Name, options.Flags, options.Mode)
+
+	_, err := fc.NextComponent().GetAttr(internal.GetAttrOptions{Name: options.Name})
+
+	// return err in case of authorization permission mismatch
+	if err != nil && err == syscall.EACCES {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
Dabnsky: OpenFile doesn't interact with any file directly. is that something that GetAttr could return to us?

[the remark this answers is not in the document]
foodprocessor: Woops, never mind!
I just examined the original OpenFile code, and it just creates the file if it didn't exist before.
We're good!

foodprocessor (approved):
Woohoo! 🥳🥳🥳

component/file_cache/file_cache_test.go
```diff
@@ -1634,8 +1645,35 @@ func (suite *fileCacheTestSuite) TestHardLimitOnSize() {
...
+	suite.assert.NoError(err)
+	suite.assert.Equal(9, n)
+	err = suite.fileCache.CloseFile(internal.CloseFileOptions{Handle: f})
+	suite.assert.NoError(err)
+
+}
+
 // In order for 'go test' to run this suite, we need to create
 // a normal test function and pass our suite to suite.Run
-func TestFileCacheTestSuite(t *testing.T) {
```
jfantinhardesty, line 1639: Why did you remove this? This is needed for the test suite to run. Right now these tests are not running in the test suite.

jfantinhardesty (changes requested):

component/file_cache/file_cache_test.go
Dabnsky: I don't know how that happened. I'll put it back

component/file_cache/file_cache_test.go
```diff
@@ -891,9 +896,9 @@ func (suite *fileCacheTestSuite) TestReadFile() {
 
 	handle, _ = suite.fileCache.OpenFile(internal.OpenFileOptions{Name: file, Mode: 0777})
 
-	d, err := suite.fileCache.ReadFile(internal.ReadFileOptions{Handle: handle})
+	d, err := suite.fileCache.ReadInBuffer(internal.ReadInBufferOptions{Handle: handle, Offset: 0, Data: data})
```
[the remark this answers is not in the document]
Dabnsky: I placed the ReadFile call back in and checked the data it returned.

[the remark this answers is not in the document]
jfantinhardesty: There still seem to be several lines where ReadFile was changed to ReadInBuffer. Could you change those back to ReadFile too?

component/file_cache/file_cache_test.go
```diff
@@ -1588,24 +1599,27 @@ func (suite *fileCacheTestSuite) TestHardLimitOnSize() {
...
+	suite.assert.False(smallHandle.Dirty())
+	err = suite.fileCache.CloseFile(internal.CloseFileOptions{Handle: smallHandle})
 	suite.assert.NoError(err)
 
+	bigHandle, err := suite.fileCache.OpenFile(internal.OpenFileOptions{Name: pathbig, Flags: os.O_RDONLY, Mode: suite.fileCache.defaultPermission})
+	suite.assert.NoError(err)
 	// try opening bigger file which shall fail due to hardlimit
-	options = internal.OpenFileOptions{Name: pathbig, Mode: 0777}
-	f, err = suite.fileCache.OpenFile(options)
+	err = suite.fileCache.downloadFile(bigHandle)
```
jfantinhardesty, line 1614: I think we should handle the issue this test brings up without downloading. They want to prevent Opening a file when the download limit is reached. The user should probably see this error on Open otherwise the reads will fail and it won't be completely transparent why. I would make a change in OpenFile to do the check for size there. There you could return the expected error.

jfantinhardesty (changes requested):

component/file_cache/file_cache_test.go
Dabnsky: This will also allow the handle to be nil as originally tested. Great suggestion. I'll get that implemented.

component/file_cache/file_cache_test.go
```diff
@@ -890,9 +896,10 @@ func (suite *fileCacheTestSuite) TestReadFile() {
 	suite.fileCache.FlushFile(internal.FlushFileOptions{Handle: handle})
 
 	handle, _ = suite.fileCache.OpenFile(internal.OpenFileOptions{Name: file, Mode: 0777})
-
+	n, err := suite.fileCache.ReadInBuffer(internal.ReadInBufferOptions{Handle: handle, Offset: 0, Data: data})
```
jfantinhardesty, line 899: This readinbuffer is here.

component/file_cache/file_cache_test.go
```diff
@@ -1473,6 +1485,10 @@ func (suite *fileCacheTestSuite) TestCachePathSymlink() {
 
 	handle, _ = suite.fileCache.OpenFile(internal.OpenFileOptions{Name: file, Mode: 0777})
 
+	n, err := suite.fileCache.ReadInBuffer(internal.ReadInBufferOptions{Handle: handle, Offset: 0, Data: data})
```
jfantinhardesty, line 1488: Another readinbuffer here.

component/file_cache/file_cache_test.go
```diff
@@ -1542,22 +1560,30 @@ func (suite *fileCacheTestSuite) TestReadFileWithRefresh() {
 	n, err := suite.fileCache.ReadInBuffer(internal.ReadInBufferOptions{Handle: f, Offset: 0, Data: data})
 	suite.assert.NoError(err)
 	suite.assert.Equal(9, n)
+	d, err := suite.fileCache.ReadFile(internal.ReadFileOptions{Handle: f})
```
jfantinhardesty, line 1563: A new readFile here.

component/file_cache/file_cache_test.go
```diff
@@ -1542,22 +1560,30 @@ func (suite *fileCacheTestSuite) TestReadFileWithRefresh() {
...
+	byteArr = []byte("test data1")
+	err = os.WriteFile(suite.fake_storage_path+"/"+path, byteArr, 0777)
 	suite.assert.NoError(err)
 	f, err = suite.fileCache.OpenFile(options)
 	suite.assert.NoError(err)
 	suite.assert.False(f.Dirty())
 	n, err = suite.fileCache.ReadInBuffer(internal.ReadInBufferOptions{Handle: f, Offset: 0, Data: data})
 	suite.assert.NoError(err)
 	suite.assert.Equal(9, n)
+	d, err = suite.fileCache.ReadFile(internal.ReadFileOptions{Handle: f})
```
jfantinhardesty, line 1579: A new readFile here.

component/file_cache/file_cache_test.go
```diff
@@ -1567,6 +1593,9 @@ func (suite *fileCacheTestSuite) TestReadFileWithRefresh() {
 	n, err = suite.fileCache.ReadInBuffer(internal.ReadInBufferOptions{Handle: f, Offset: 0, Data: data})
 	suite.assert.NoError(err)
 	suite.assert.Equal(15, n)
+	d, err = suite.fileCache.ReadFile(internal.ReadFileOptions{Handle: f})
```
jfantinhardesty, line 1596: A new readFile here.

jfantinhardesty (changes requested):

jfantinhardesty (approved):

PR #248, opened 2024-06-12
Improve s3 error messages

jfantinhardesty:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Improve S3 error handling code. This adds additional errors for invalid credentials, endpoint, bucket, and regions along with specific portions of the error that will be parsed by our plugin.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/s3storage/client.go
+++ b/component/s3storage/client.go
@@ -33,6 +33,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"net/http"
 	"net/url"
 	"os"
 	"strings"
@@ -45,6 +46,7 @@ import (
 	"github.com/Seagate/cloudfuse/internal/stats_manager"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awsHttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/s3"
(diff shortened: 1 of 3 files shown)


Dabnsky (approved):
Looks pretty clear and good to me. Great job!

component/s3storage/client.go
```diff
@@ -45,10 +46,12 @@ import (
 	"github.com/Seagate/cloudfuse/internal/stats_manager"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awsHttp "github.com/aws/aws-sdk-go-v2/aws/transport/http"
```
Ferelith-maker, line 49: What does awsHTTP mean on this line?

jfantinhardesty: Yeah, the awsHttp represents the name of the library. It would normally be named http, but net/http is also imported so their would be a name collision. So this just renames the one for aws.

component/s3storage/client.go
```diff
@@ -135,20 +149,67 @@ func (cl *Client) Configure(cfg Config) error {
...
 
+	// If endpoint is localhost, then running locally so ignore region check
+	isLocalHost := strings.Contains(cl.Config.authConfig.Endpoint, "localhost")
+
+	// Check that region matches region in endpoint
+	// This check is after endpoint errors so endpoint can be validated first
+	if !strings.Contains(cl.Config.authConfig.Endpoint, cl.Config.authConfig.Region) && !isLocalHost {
+		log.Err("Endpoint in region does not match provided endpoint.")
+		return errRegionMismatch
+	}
```
foodprocessor, line 192: This would break when the endpoint has no region.
Can you use the error coming back from the cloud storage provider to identify this case instead?

component/s3storage/client.go
```diff
@@ -135,20 +149,67 @@ func (cl *Client) Configure(cfg Config) error {
...
 	// if no bucket-name was set, default to the first bucket in the list
-	if cl.Config.authConfig.BucketName == "" && len(bucketList) > 0 {
-		cl.Config.authConfig.BucketName = bucketList[0]
-		log.Warn("Client::Configure : Bucket defaulted to first listed bucket: %s", bucketList[0])
+	if cl.Config.authConfig.BucketName == "" {
+		if len(bucketList) > 0 {
+			cl.Config.authConfig.BucketName = bucketList[0]
+			log.Warn("Client::Configure : Bucket defaulted to first listed bucket: %s", bucketList[0])
+		} else {
+			log.Err("Client::Configure : Error no bucket exists in account. Here's why: %v", err)
```
foodprocessor, line 200: The "Here's why" seems strange in this case.

foodprocessor (changes requested):
This is great! Thank you!

component/s3storage/client.go
jfantinhardesty: Good point, I moved this after we make a request.

component/s3storage/client.go
jfantinhardesty: Removed.

foodprocessor (approved):
👌

PR #263, opened 2024-06-26
Network mount name

Dabnsky:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

bucket name now displayed for network mount on windows.
### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
foreground set to false does not implement this change.
- Closes #

### DIFF
--- a/component/libfuse/libfuse2_handler.go
+++ b/component/libfuse/libfuse2_handler.go
@@ -183,7 +183,9 @@ func (lf *Libfuse) initFuse() error {
 				kindStorage = "cloud"
 				log.Err("initFuse : Failed to unmarshal s3storage.bucket-name")
 			}
-
+		} else {
+			nameStorage = "default"
+			kindStorage = "cloud"
 		}
 
 		volumePrefix := fmt.Sprintf("--VolumePrefix=\\%s\\%s", kindStorage, nameStorage)
--- a/component/libfuse/libfuse2_handler.go
+++ b/component/libfuse/libfuse2_handler.go
@@ -173,7 +173,7 @@ func (lf *Libfuse) initFuse() error {
 			}
 		}
 
-		volumePrefix := fmt.Sprintf("--VolumePrefix=\\bucket: %s\\share", bucketName)
+		volumePrefix := fmt.Sprintf("--VolumePrefix=\\bucket\\%s", bucketName)
 		opts = append(opts, volumePrefix)
 	}
 
--- a/component/libfuse/libfuse2_handler.go
+++ b/component/libfuse/libfuse2_handler.go
@@ -183,9 +183,7 @@ func (lf *Libfuse) initFuse() error {
 				kindStorage = "cloud"
 				log.Err("initFuse : Failed to unmarshal s3storage.bucket-name")
 			}
-		} else {
-			nameStorage = "default"
-			kindStorage = "cloud"
+
 		}
 
 		volumePrefix := fmt.Sprintf("--VolumePrefix=\\%s\\%s", kindStorage, nameStorage)
--- a/component/libfuse/libfuse2_handler.go
+++ b/component/libfuse/libfuse2_handler.go
@@ -36,6 +36,7 @@ import (
 	"time"
 
 	"github.com/Seagate/cloudfuse/common"
+	"github.com/Seagate/cloudfuse/common/config"
 	"github.com/Seagate/cloudfuse/common/log"
 	"github.com/Seagate/cloudfuse/internal"
 	"github.com/Seagate/cloudfuse/internal/handlemap"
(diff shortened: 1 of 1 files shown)


component/libfuse/libfuse2_handler.go
```diff
@@ -160,7 +161,20 @@ func (lf *Libfuse) initFuse() error {
 	if runtime.GOOS == "windows" && lf.networkShare && common.IsDriveLetter(lf.mountPath) {
 		// TODO: We can support any type of valid network share path so this path could
 		// be configurable for the config file. But this is a good default.
-		opts = append(opts, "--VolumePrefix=\\server\\share")
+
+		// by default bucketName will be blank
+		bucketName := ""
+		// Borrow bucket-name string from attribute cache
+		if config.IsSet("s3storage.bucket-name") {
```
jfantinhardesty, line 168: Could you do the same thing but for Azure here?

foodprocessor: Good point. You want both, depending on which one we're using, right? Does that mean we should check the components section of the config to see which cloud we're using?
This might be a bit tricky. I'm not sure if `config.IsSet("components.s3storage")` (or `azstorage`) will work...

foodprocessor (changes requested):
Making it work for both cloud storage providers sounds like a good requirement.
As usual, I was not thinking of Azure either!

component/libfuse/libfuse2_handler.go
Dabnsky: I'm afraid the config component section is the only way this can be checked. Doesn't the mount fail if that information is missing in the config anyway?

foodprocessor: There is code that writes a default components section when it is missing from the config. That being said, it's fine to do something that _usually_ works, and base it on the config.

Dabnsky: I was wrong that the components section is the only place. Just checking the s3storage or azstorage section of the config is good enough and the mound fails without a container or bucket specified in the config.

component/libfuse/libfuse2_handler.go
```diff
@@ -160,7 +161,33 @@ func (lf *Libfuse) initFuse() error {
...
+		var kindStorage string
+		// Borrow bucket-name string from attribute cache
+		if config.IsSet("s3storage.bucket-name") {
+
+			err := config.UnmarshalKey("s3storage.bucket-name", &nameStorage)
+			kindStorage = "bucket"
+			if err != nil {
+				nameStorage = "default"
+				kindStorage = "Cloud Storage"
+				log.Err("ParseAndReadDynamicConfig : Failed to unmarshal s3storage.bucket-name")
```
jfantinhardesty, line 176: Could you change the error here to match the function.

component/libfuse/libfuse2_handler.go
```diff
@@ -160,7 +161,33 @@ func (lf *Libfuse) initFuse() error {
...
+			}
+		} else if config.IsSet("azstorage.container") {
+			err := config.UnmarshalKey("azstorage.container", &nameStorage)
+			kindStorage = "container"
+			if err != nil {
+				nameStorage = "default"
+				kindStorage = "Cloud Storage"
+				log.Err("ParseAndReadDynamicConfig : Failed to unmarshal s3storage.bucket-name")
```
jfantinhardesty, line 184: Same thing here but should also be for azstorage.container.

component/libfuse/libfuse2_handler.go
```diff
@@ -160,7 +161,33 @@ func (lf *Libfuse) initFuse() error {
...
+			}
+
+		}
```
jfantinhardesty, line 187: You need an else case here in case they are using our feature where we select the first bucket listed. (This is how our plugin would work.) If it doesn't recognize a container or a bucket then the mount will fail.

component/libfuse/libfuse2_handler.go
```diff
@@ -160,7 +161,33 @@ func (lf *Libfuse) initFuse() error {
...
+				nameStorage = "default"
+				kindStorage = "Cloud Storage"
+				log.Err("ParseAndReadDynamicConfig : Failed to unmarshal s3storage.bucket-name")
+			}
+		} else if config.IsSet("azstorage.container") {
+			err := config.UnmarshalKey("azstorage.container", &nameStorage)
+			kindStorage = "container"
+			if err != nil {
+				nameStorage = "default"
+				kindStorage = "Cloud Storage"
```
jfantinhardesty, line 183: I wouldn't do a space here, instead maybe just do "cloud"

jfantinhardesty (changes requested):

component/libfuse/libfuse2_handler.go
Dabnsky: Are you suggesting to fetch the first bucket list in the else case? or just name the network mount as "cloud default" in that case?

jfantinhardesty (approved):

foodprocessor (approved):

PR #287, opened 2024-07-17
Async Fault Tolerance

ahuynh1289:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Added cloud-down fault tolerance so that file changes are preserved while cloud connection is down, and will be updated in the cloud once connection comes back. Still need to add a mechanism that evicts files from the cache if the file was updated while the cloud was down and cache timeout has been exceeded.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #


block.sh
foodprocessor, line 1: This probably shouldn't be at the root of the repo. Some poor soul may run this script and be very unhappy. Maybe `block.sh` and `connect.sh` belong in the `test-scripts` folder? `start.sh` could move there too.

Dabnsky (changes requested):
Great Job! though I think I'll need a tour to really understand what's going on.

component/attr_cache/attr_cache.go
```diff
@@ -341,6 +344,11 @@ func (ac *AttrCache) CreateDir(options internal.CreateDirOptions) error {
 			exists:   true,
 			cachedAt: time.Now(),
 		})
+		if newDirAttrCacheItem != nil {
```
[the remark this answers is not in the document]
ahuynh1289: I initially added that there to help me debug and make sure that createDir was working as intended. Just removed that

component/file_cache/file_cache.go
```diff
@@ -160,6 +175,7 @@ func (c *FileCache) Priority() internal.ComponentPriority {
 //	this shall not block the call otherwise pipeline will not start
 func (c *FileCache) Start(ctx context.Context) error {
 	log.Trace("Starting component : %s", c.Name())
+	//c.closeSignal = make(chan int)
```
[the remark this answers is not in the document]
ahuynh1289: I had that there to help indicate to the async thread whether or not it could stop running. I ran into some issues with it because it was ending the thread early, so I just commented it out with the plan to revisit it at a later time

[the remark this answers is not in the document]
ahuynh1289: I just rely on the go compilier to cleanup running threads when we end the program, but I'm not sure if that's something we want in the long term

component/file_cache/file_cache.go
```diff
@@ -466,19 +490,70 @@ func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.O
...
+					attrs = append(attrs[:i], attrs[i+1:]...)
+				}
+				//TODO: Add support for rename directory
+				if val.operation == "RenameDir" {
+					renameOptions := val.options.(internal.RenameDirOptions)
+					attrs[i].Name = renameOptions.Dst //this just does a rename but we need to make sure file contents are also transferred
+
+				}
+			}
+			return true
```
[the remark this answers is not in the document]
ahuynh1289: This is just part of the syntax when using range on a sync.map. Return true indicates we can move to the next element in the map

component/file_cache/file_cache_linux_test.go
```diff
@@ -113,8 +113,12 @@ func (suite *fileCacheLinuxTestSuite) TestChmodNotInCache() {
 	defer suite.cleanupTest()
 	// Setup
 	path := "file33"
+	defaultConfig := fmt.Sprintf("file_cache:\n  path: %s\n  offload-io: true\n  timeout-sec: 1\n\nloopbackfs:\n  path: %s", suite.cache_path, suite.fake_storage_path)
+	suite.setupTestHelper(defaultConfig)
 	handle, _ := suite.fileCache.CreateFile(internal.CreateFileOptions{Name: path, Mode: 0777})
+	time.Sleep(time.Millisecond)
```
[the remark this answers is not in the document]
ahuynh1289: Since the new design is asynchronous, we need to give the async thread some time to complete its work before doing the next file operation. In this case, it tests chmod, but we give short timeout to ensure that the thread had enough time to actually create the file

component/file_cache/file_cache_test.go
```diff
@@ -301,16 +335,39 @@ func (suite *fileCacheTestSuite) TestCreateDir() {
 	path := "a"
 	options := internal.CreateDirOptions{Name: path}
 	err := suite.fileCache.CreateDir(options)
+	time.Sleep(time.Millisecond)
```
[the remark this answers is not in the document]
ahuynh1289: Same reason as above, I added the timeout to give the async thread time to finish executing.

component/file_cache/file_cache_linux_test.go
```diff
@@ -286,8 +315,7 @@ func (suite *fileCacheLinuxTestSuite) TestChownCase2() {
 	owner := os.Getuid()
 	group := os.Getgid()
 	err := suite.fileCache.Chown(internal.ChownOptions{Name: path, Owner: owner, Group: group})
-	suite.assert.Error(err)
-	suite.assert.Equal(syscall.EIO, err)
+	time.Sleep(time.Millisecond)
```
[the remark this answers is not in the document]
ahuynh1289: In the old implementation of Chown, it would first change owners in the cloud and then do it locally. Because this is a case 2 test, the file is only in local storage, so it expects an error back because the file does not exist in the cloud. I removed the check because in the new implementation, all local file operations are serviced in one thread, and all cloud operations are serviced on another. We wouldn't get an error back from chown because it successfully changed the owner locally. I can check for no error if that's fine

component/file_cache/file_cache.go
```diff
@@ -891,6 +1026,35 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
...
+
+	_, loaded := fc.fileOps.LoadOrStore(options.Name, nextAttr)
+
+	if loaded {
+
+		fc.fileOps.Delete(options.Name)
+		fc.fileOps.Store(options.Name, nextAttr)
+	}
+	fc.asyncSignal.TryLock() // Make sure we don't unlock a mutex that is not locked
+	fc.asyncSignal.Unlock()  // Signal to async thread to do work
```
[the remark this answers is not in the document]
ahuynh1289: Yeah it's a bit of a weird workaround to try and stall the async thread from spinning until there's work to be done. Normally a mutex indicates that we want to lock other threads from accessing a critical section, but in this use case we are using it as a signal that we can begin to do work. I used a mutex because they automatically put the thread to sleep and awake when unlocked, so we don't need to do any sleep management on our own. I can go more in depth with the specific implementation

component/file_cache/lru_policy.go
```diff
@@ -387,9 +426,28 @@ func (p *lruPolicy) deleteExpiredNodes() {
 	}
 
 	for ; node != nil && count < p.maxEviction; node = node.next {
-		delItems = append(delItems, node)
-		node.deleted = true
-		count++
+		p.fileOps.Range(func(key, value interface{}) bool {
```
jfantinhardesty, line 429: This Range function seems to be replicated a few times. Do you think it would be worth having a separate function for that?

component/file_cache/file_cache_linux.go
```diff
@@ -75,6 +77,25 @@ func (fc *FileCache) isDownloadRequired(localPath string, blobPath string, flock
...
 		downloadRequired = true
 	}
+	//check async map and automatically set downloadRequired false if entry is matched
+	//When cloud is down, local data will always be most up to date compared to cloud, download is not required
+	//Async map stores file entries that have been changed when cloud is down
+
+	//case is no record of file in cloud
+	//does not exist locally
+	//this does not matter because isDownloadRequired is called by OpenFile, which means file already exists somewhere
+	fc.fileOps.Range(func(key, value interface{}) bool {
```
jfantinhardesty, line 87: Could you cleanup some of the empty lines here.

component/file_cache/file_cache_async.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	return nil
+}
+
+func (fc *FileCache) asyncChown(options internal.ChownOptions) error {
+
+	// flushFilePath := FlushFileAbstraction{}
+	// flushFilePath.Name = options.Name
+
+	// if !fc.createEmptyFile {
```
jfantinhardesty, line 304: Does this commented out section need to be implemented later. Like there could still be an issue with empty file?

component/file_cache/file_cache_async.go
```diff
@@ -0,0 +1,331 @@
...
+
+	localPath := common.JoinUnixFilepath(fc.tmpPath, options.Name)
+	uploadHandle, err := common.Open(localPath)
+
+	if err != nil {
+		log.Err("FileCache::FlushFile : error [unable to open upload handle] %s [%s]", options.Name, err.Error())
+		return err
+	}
+
+	err = fc.NextComponent().CopyFromFile(
```
jfantinhardesty, line 146: Should this also call the next components FlushFile?

component/file_cache/file_cache_async.go
```diff
@@ -0,0 +1,331 @@
...
+							tries++ //failed op, increase timeout duration
+						}
+
+					}
+				}
+				return true
+			})
+		}
+
+		//check channel to end thread here
```
jfantinhardesty, line 127: Should this be uncommented?

component/file_cache/file_cache.go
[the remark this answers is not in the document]
jfantinhardesty: Is there a reason you chose to do it this way, rather than with a channel?

component/file_cache/file_cache.go
```diff
@@ -750,14 +882,7 @@ func (fc *FileCache) downloadFile(handle *handlemap.Handle) error {
 			fileSize = int64(attr.Size)
 		}
 
-		if fileExists {
```
jfantinhardesty, line 753: Could you explain this change? I believe this section is intended to get the latest copy from the cloud when download file is called so the local copy is first cleared.

component/file_cache/file_cache.go
```diff
@@ -466,19 +490,70 @@ func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.O
...
+
+			if val.operation == "RenameFile" || val.operation == "RenameDir" { //for rename operations, only grab the src file/dir to compare
+
+				keyString = keyString[:strings.IndexByte(keyString, ',')]
+
+			}
+
+			if attr.Name == keyString {
+				//remove file listing if there is a deleteFile operation
+				if val.operation == "DeleteFile" {
```
jfantinhardesty, line 524: Could you use a switch statement here, it will look a bit cleaner to read.

component/file_cache/file_cache.go
```diff
@@ -73,9 +75,22 @@ type FileCache struct {
...
 	hardLimit         bool
 	diskHighWaterMark float64
+	lazyWrite         bool
+	fileCloseOpt      sync.WaitGroup
 
-	lazyWrite    bool
-	fileCloseOpt sync.WaitGroup
+	fileOps     sync.Map //we want fileOps to store the operations to perform on the file if the cloud is down. Key is file name, value is operation to do
+	asyncSignal sync.Mutex
+	//closeSignal chan int
```
Ferelith-maker, line 83: If closeSignal isn't being used, this comment can be taken out

component/file_cache/file_cache.go
```diff
@@ -466,19 +490,70 @@ func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.O
...
 	}
 
 	// Get files from local cache
 	localPath := common.JoinUnixFilepath(fc.tmpPath, options.Name)
-	dirents, err := os.ReadDir(localPath)
-	if err != nil {
-		return attrs, token, nil
-	}
+	dirents, _ := os.ReadDir(localPath)
+	// if err != nil {
```
Ferelith-maker, line 551: Is this comment section meant for debugging and will be removed later?

component/file_cache/file_cache_linux_test.go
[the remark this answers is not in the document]
Dabnsky: That's fine. in that case, you would use suite.assert.NoError(err)

component/file_cache/file_cache.go
```diff
@@ -466,19 +490,70 @@ func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.O
...
+			if attr.Name == keyString {
+				//remove file listing if there is a deleteFile operation
+				if val.operation == "DeleteFile" {
+					attrs = append(attrs[:i], attrs[i+1:]...)
+				}
+				//Change source file name to dst file name
+				if val.operation == "RenameFile" {
+					renameOptions := val.options.(internal.RenameFileOptions)
+					attrs[i].Name = renameOptions.Dst //this just does a rename but we need to make sure file contents are also transferred
+				}
```
foodprocessor, line 531: Since rename can also move a file (or folder), you might need to check if the file still belongs in the this directory listing after its rename.

component/file_cache/file_cache.go
```diff
@@ -466,19 +490,70 @@ func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.O
...
+				if val.operation == "RenameDir" {
+					renameOptions := val.options.(internal.RenameDirOptions)
+					attrs[i].Name = renameOptions.Dst //this just does a rename but we need to make sure file contents are also transferred
+
+				}
+			}
+			return true
+		})
+		i++
 	}
```
foodprocessor, line 546: This code alters or removes entries returned from the attribute cache which overlap with entries in `fc.fileOps`.
But I think there are cases when we need to add entries from `fc.fileOps` which are not in the listing from the attribute cache.

foodprocessor (changes requested):
I think I see an issue in StreamDir which is probably causing some of the e2e test failures.

component/s3storage/client.go
```diff
@@ -826,6 +841,9 @@ func (cl *Client) Write(options internal.WriteFileOptions) error {
 			err = cl.ReadInBuffer(name, fileOffsets.BlockList[index].StartIndex, oldDataSize, oldDataBuffer)
 			if err != nil {
 				log.Err("BlockBlob::Write : Failed to read data in buffer %s [%s]", name, err.Error())
+				if err != syscall.ENOENT {
```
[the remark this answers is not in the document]
ahuynh1289: I believe these changes were made by Michael!

component/s3storage/client.go
```diff
@@ -128,6 +129,9 @@ func (cl *Client) Configure(cfg Config) error {
 		config.WithCredentialsProvider(credentialsProvider),
 		config.WithAppID(UserAgent()),
 		config.WithRegion(cl.Config.authConfig.Region),
+		config.WithRetryer(func() aws.Retryer {
```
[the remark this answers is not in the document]
ahuynh1289: Yes this was mainly to speed up testing, and I think we've been slowly increasing it as we do more testing because some operations end up taking a while to get to the cloud. We could probably return it back to normal, but it can take up to 20-30 seconds on an operation with full attempts and timeout

component/file_cache/cache_policy.go
```diff
@@ -97,6 +98,8 @@ func getUsagePercentage(path string, maxSize float64) float64 {
 // Delete a given file
 func deleteFile(name string) error {
 	log.Debug("cachePolicy::deleteFile : attempting to delete %s", name)
+	//name is grabbing the whole local path, but the sync map only stores the file/dir name not the whole path
+	//rework name to only store the file name rather than full path
```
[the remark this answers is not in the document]
ahuynh1289: I just had a small confusion on whether or not the cache stored full file paths, or just the file name. I removed the comment

component/file_cache/file_cache_linux.go
ahuynh1289: Done

component/file_cache/file_cache_async.go
ahuynh1289: I think we resolved this already! Deleted the comments

component/file_cache/file_cache_async.go
ahuynh1289: In the original design, fileCache flushFile used copyFromFile so I just did the same thing here

component/file_cache/file_cache_async.go
ahuynh1289: I tried to implement a shut down mechanism with this for the async thread, but it was very buggy. I'll remove it and leave a comment to revisit it at some point

component/file_cache/file_cache.go
ahuynh1289: Yes that is correct, but I remember we were running into an issue with this deleting files when the cloud was down, so the local copy of the file gets deleted but there's no cloud to copy from. When retrieving the cloud copy of the file it will just overwrite the local copy so we don't need to delete. Instead, just check for file existence and make the file if it doesn't exist already.

component/file_cache/file_cache.go
ahuynh1289: done

component/file_cache/file_cache.go
ahuynh1289: Yes I have removed it

component/file_cache/file_cache_async.go
```diff
@@ -0,0 +1,331 @@
...
+			File: uploadHandle,
+		})
+
+	uploadHandle.Close()
+	if err != nil {
+		log.Err("FileCache::FlushFile : %s upload failed [%s]", options.Name, err.Error())
+		return err
+	}
+
+	//use miss list to update mode, take it out of flush file
```
Ferelith-maker, line 158: Is there code attached to this comment that was meant to be fixed later?

component/file_cache/file_cache_async.go
```diff
@@ -0,0 +1,331 @@
...
+		err = fc.NextComponent().RenameFile(options)
+		err = fc.validateStorageError(options.Src, err, "RenameFile", false)
+		if err != nil {
+			log.Err("FileCache::RenameFile : %s failed to rename file [%s]", options.Src, err.Error())
+			return err
+		}
+		return nil
+
+	}
+	// err = fc.NextComponent().RenameFile(options)
```
Ferelith-maker, line 203: Is this commented code section for debug?

component/file_cache/file_cache_async.go
ahuynh1289: This was for an outdated idea, the comment just didn't get removed. I deleted it!

component/file_cache/file_cache_async.go
ahuynh1289: It was just a section of code I needed, but in a different place. Removed that comment!

foodprocessor:
We can continue to reference this PR as we develop network fault tolerance, but I'm going to close it to get it off the board for now.
When it's time to merge network fault tolerance, we can put those commits on this branch and re-open this PR for the merge in order to give credit to @ahuynh1289 for the first design and implementation for this feature.

PR #312, opened 2024-08-30
Add display capacity option

foodprocessor:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### Checklist

- [ ] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -56,11 +56,11 @@ const (
 	DefaultDirectoryPermissionBits  os.FileMode = 0775
 	DefaultAllowOtherPermissionBits os.FileMode = 0777
 
-	MbToBytes       = 1024 * 1024
-	GbToBytes       = 1024 * 1024 * 1024
-	TbToBytes       = 1024 * 1024 * 1024 * 1024
-	DefaultCapacity = TbToBytes
-	CfuseStats      = "cloudfuse_stats"
+	MbToBytes         = 1024 * 1024
+	GbToBytes         = 1024 * 1024 * 1024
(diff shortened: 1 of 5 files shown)


common/types.go
```diff
@@ -56,10 +56,11 @@ const (
...
 	DefaultAllowOtherPermissionBits os.FileMode = 0777
 
-	MbToBytes  = 1024 * 1024
-	GbToBytes  = 1024 * 1024 * 1024
-	TbToBytes  = 1024 * 1024 * 1024 * 1024
-	CfuseStats = "cloudfuse_stats"
+	MbToBytes       = 1024 * 1024
+	GbToBytes       = 1024 * 1024 * 1024
+	TbToBytes       = 1024 * 1024 * 1024 * 1024
+	DefaultCapacity = TbToBytes
```
jfantinhardesty, line 62: I think it might be a bit cleaner to have this as DefaultCapacityMB and have it in MB here so we don't have to do the division to MB in case this changes in the future.

jfantinhardesty (approved):

common/types.go
foodprocessor: Like this?
`DefaultCapacityMb = TbToBytes / MbToBytes`

jfantinhardesty: I think either is fine, but the first one might be a bit clearer that it is a terabyte.

PR #314, opened 2024-09-03
Refactor gui settings

Ferelith-maker:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Remove the code that uses QSettings for config settings and use an internal dictionary to pass values between widgets. 
Refactor the MainFactor to use config functions class
### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


gui/common_qt_functions.py
```diff
@@ -174,40 +176,111 @@ def setComponentSettings(self):
...
+                self.settings[option]=tempDict
+            else:
+                self.settings[option] = dictForConfigs[option]
+
+    def getConfigs(self,useDefault=False):
+        workingDir = self.getWorkingDir()
+        if useDefault:
+            # Use programmed defaults
+            defaultSettingsManager.setAllDefaultSettings(self,self.settings)
+            configs = self.settings
```
foodprocessor, line 245: Since `getConfigs` is not part of the `defaultSettingsManager` class, `self` means an instance of `customConfigFunctions` here, so does this call to `setAllDefaultSettings` create a separate dictionary and write to the new `settings` field in `customConfigFunctions`?
Or is `self.settings` defined somewhere else in `customConfigFunctions`?

foodprocessor (commented):
I'm in over my head!

gui/common_qt_functions.py
Ferelith-maker: Fixed

Dabnsky (approved):

foodprocessor (approved):
Oh my god inheritance is so hard to understand in Python! I have a headache!
But I'm pretty sure this is right! 🎉

gui/common_qt_functions.py
```diff
@@ -248,74 +318,16 @@ def closeEvent(self, event):
         elif ret == QtWidgets.QMessageBox.Save:
             # Insert all settings to yaml file
             self.exitWindowCleanup()
-            if self.writeConfigFile():
+            self.updateSettingsFromUIChoices()
+            if self.writeConfigFile(self.settings):
```
foodprocessor, line 322: I'm pretty sure this is correct, because `closeEvent` is called with `self` set to the widget, so the widget's `self.settings` would be accessible here.
But I'm writing this here, because it's the one place where I'm not completely sure.

PR #333, opened 2024-09-28
Minor improvements in file cache

foodprocessor:

In file cache test suite:

* Re-enable tests that were flaky due to race conditions that are fixed now
* Shorten a couple test sleep times
* Remove unnecessary OpenFile when timeout is non-zero
* Remove redundant call to setupTestHelper with config identical to test default config

In lru_policy:

* Remove unused const
* Simplify cacheTimeoutMonitor code
  * (calling time.Tick with a zero interval already returns nil, so no need to check if the interval is zero)
* Add comments to updateMarker to explain linked list code

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### Checklist

- [ ] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/file_cache/file_cache_test.go
+++ b/component/file_cache/file_cache_test.go
@@ -1369,6 +1369,7 @@ func (suite *fileCacheTestSuite) TestRenameFileAndCacheCleanup() {
 	err := suite.fileCache.RenameFile(internal.RenameFileOptions{Src: src, Dst: dst})
 	suite.assert.NoError(err)
 	// Path in fake storage and file cache should be updated
+	suite.assert.False(suite.fileCache.policy.IsCached(filepath.Join(suite.cache_path, src)))
 	suite.assert.NoFileExists(suite.cache_path + "/" + src)        // Src does not exist
 	suite.assert.FileExists(suite.cache_path + "/" + dst)          // Dst shall exists in cache
 	suite.assert.NoFileExists(suite.fake_storage_path + "/" + src) // Src does not exist
@@ -1399,6 +1400,7 @@ func (suite *fileCacheTestSuite) TestRenameFileAndCacheCleanupWithNoTimeout() {
 	err := suite.fileCache.RenameFile(internal.RenameFileOptions{Src: src, Dst: dst})
 	suite.assert.NoError(err)
 	// Path in fake storage and file cache should be updated
+	suite.assert.False(suite.fileCache.policy.IsCached(filepath.Join(suite.cache_path, src)))
 	suite.assert.NoFileExists(suite.cache_path + "/" + src)        // Src does not exist
 	suite.assert.FileExists(suite.cache_path + "/" + dst)          // Dst shall exists in cache
 	suite.assert.NoFileExists(suite.fake_storage_path + "/" + src) // Src does not exist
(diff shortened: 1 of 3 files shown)


component/file_cache/file_cache_test.go
```diff
@@ -1252,9 +1247,7 @@ func (suite *fileCacheTestSuite) TestGetAttrCase4() {
 		time.Sleep(10 * time.Millisecond)
 		_, err = os.Stat(filepath.Join(suite.cache_path, file))
 	}
-	// TODO: why is check test flaky (on both platforms)?
```
Ferelith-maker, line 1255: Is this test no longer flaky?

component/file_cache/lru_policy.go
```diff
@@ -189,8 +183,6 @@ func (p *lruPolicy) CachePurge(name string) {
 	p.deleteEvent <- name
 }
 
-// Due to a race condition, this may return a false positive,
```
Ferelith-maker, line 192: Where is the fix for the race condition mentioned in the comments?

component/file_cache/file_cache_test.go
foodprocessor: Honestly, I'm not sure. I don't understand the root cause for this one, and I don't have a smoking-gun explanation for why my recent changes would have fixed this issue. I just had a hunch that this might not come up again and thought it would be good to try it.

component/file_cache/lru_policy.go
foodprocessor: The code that fixes this race condition is in `asyncCacheValid()`, where it now checks if a file is in the `nodeMap` before calling `cacheValidate()`.
The comment there explains the bug and the fix.

component/file_cache/file_cache_test.go
```diff
@@ -1377,52 +1369,45 @@ func (suite *fileCacheTestSuite) TestRenameFileAndCacheCleanup() {
...
+
+	// Dst _might_ exist in cache, and the rename should be complete in fake storage
 	suite.assert.NoFileExists(suite.fake_storage_path + "/" + src) // Src does not exist
 	suite.assert.FileExists(suite.fake_storage_path + "/" + dst)   // Dst does exist
 
-	suite.fileCache.CloseFile(internal.CloseFileOptions{Handle: openHandle})
-
-	time.Sleep(100 * time.Millisecond)                      // Wait for the cache cleanup to occur
-	suite.assert.NoFileExists(suite.cache_path + "/" + dst) // Dst shall not exists in cache
+	time.Sleep(100 * time.Millisecond) // Wait for the cache cleanup to occur
```
Dabnsky, line 1405: I no longer see the config being defined with the `timeout-sec`. is there a default i'm missing?

component/file_cache/lru_policy.go
```diff
@@ -74,9 +74,6 @@ type lruPolicy struct {
 }
 
 const (
-	// Check for file expiry in below number of seconds
-	CacheTimeoutCheckInterval = 5
```
Dabnsky, line 78: by taking this out, is there a new default that's taking place? was this magic number a significant factor in the race conditions? or was it just overkill?

Dabnsky (approved):
looks pretty good to me. Great job

component/file_cache/file_cache_test.go
foodprocessor: The default config set in `SetupTest()` [on line 101](https://github.com/Seagate/cloudfuse/blob/minor-improvements-to-filecache/component/file_cache/file_cache_test.go#L101) is identical to the custom config that was used in this test. So I just dropped the custom config here since it was redundant.

component/file_cache/lru_policy.go
foodprocessor: This constant was just never used. So I cleaned it up.
I have no idea what it was intended for.

Ferelith-maker (approved):
Your comments make sense. Good job

PR #338, opened 2024-10-10
Cloudfuse service linux cmd

Dabnsky:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This adds the following commands available to the cloudfuse executable:
cloudfuse service install
cloudfuse service uninstall 

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/cmd/service_linux.go
+++ b/cmd/service_linux.go
@@ -80,7 +80,7 @@ var installCmd = &cobra.Command{
 			return fmt.Errorf("error collecting data from cloudfuse.service file due to the following error: [%s]", err)
 		}
 
-		//TODO: set a default mount and config path when stubbed examples are found in the file.
+		// set a default mount and config path when stubbed examples are found in the file.
 		mountPath := serviceData["MoutingPoint"]
 		configPath := serviceData["ConfigFile"]
 		if mountPath == "/path/to/mounting/point" {
@@ -114,7 +114,7 @@ var installCmd = &cobra.Command{
 		}
 		if !foundUser {
 			//create the user
-			userAddCmd := exec.Command("useradd", "-m", user)
+			userAddCmd := exec.Command("sudo", "useradd", "-m", user)
 			err := userAddCmd.Run()
 			if err != nil {
 				return fmt.Errorf("failed to create user due to following error: [%s]", err.Error())
@@ -123,7 +123,7 @@ var installCmd = &cobra.Command{
 
 		// 3. copy the cloudfuse.service file to /etc/systemd/system
 
-		copyFileCmd := exec.Command("cp", "./setup/cloudfuse.service", "/etc/systemd/system")
+		copyFileCmd := exec.Command("sudo", "cp", "./setup/cloudfuse.service", "/etc/systemd/system")
 		err = copyFileCmd.Run()
 		if err != nil {
 			return fmt.Errorf("failed to copy cloudfuse.service file to /etc/systemd/system due to following error: [%s]", err.Error())
@@ -131,7 +131,7 @@ var installCmd = &cobra.Command{
 
 		// 4. run systemctl daemon-reload
 
-		systemctlCmd := exec.Command("systemctl", "daemon-reload")
+		systemctlCmd := exec.Command("sudo", "systemctl", "daemon-reload")
 		err = systemctlCmd.Run()
 		if err != nil {
 			return fmt.Errorf("failed to run 'systemctl daemon-reload' command due to following error: [%s]", err.Error())
@@ -216,7 +216,7 @@ func modifySericeFile(path string, curDir string) error {
 	}
 	if strings.Contains(path, "mounting") {
 		defaultMountPath = fmt.Sprintf("/home/%s/cloudfuseMount", strings.ToLower(usr.Name))
-		fmt.Printf("creating mount folder in %s", defaultMountPath)
+		fmt.Printf("creating mount folder in %s \n", defaultMountPath)
 
 		//Create default mount directory
 		userAddCmd := exec.Command("mkdir", defaultMountPath)
--- a/cmd/service_linux.go
+++ b/cmd/service_linux.go
@@ -1,0 +1,100 @@
+//go:build linux
+
+/*
+   Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+
+   Copyright © 2023-2024 Seagate Technology LLC and/or its Affiliates
+
+   Permission is hereby granted, free of charge, to any person obtaining a copy
+   of this software and associated documentation files (the "Software"), to deal
+   in the Software without restriction, including without limitation the rights
+   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+   copies of the Software, and to permit persons to whom the Software is
+   furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in all
+   copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+   AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+*/
+
+package cmd
+
+import (
+	"errors"
+
+	"github.com/spf13/cobra"
+)
+
+type serviceOptions struct {
+	ConfigFile string
+	MountPath  string
+}
+
+const SvcName = "cloudfuse.service"
+
+var servOpts serviceOptions
+
+// Section defining all the command that we have in secure feature
+var serviceCmd = &cobra.Command{
+	Use:               "service",
+	Short:             "Manage cloudfuse startup process on Linux",
+	Long:              "Manage cloudfuse startup process on Linux",
+	SuggestFor:        []string{"ser", "serv"},
+	Example:           "cloudfuse service install",
+	FlagErrorHandling: cobra.ExitOnError,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return errors.New("missing command options\n\nDid you mean this?\n\tcloudfuse service install\n\nRun 'cloudfuse service --help' for usage")
+	},
+}
+
+var installCmd = &cobra.Command{
+	Use:               "install",
+	Short:             "Installs the startup process for Cloudfuse. Requires elevated permissions.",
+	Long:              "Installs the startup process for Cloudfuse which remounts any active previously active mounts on startup. elevated permissions.",
+	SuggestFor:        []string{"ins", "inst"},
+	Example:           "cloudfuse service install",
+	FlagErrorHandling: cobra.ExitOnError,
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		// if err != nil {
+		// 	return fmt.Errorf("unable to determine location of cloudfuse binary [%s]", err.Error())
+		// }
+
+		// 1. get the cloudfuse.service file from the setup folder and verify its contents (the mount path and config path is valid)
+		// 2. retrieve the user account from cloudfuse.service file and make it if it doesn't exist
+		// 3. copy the cloudfuse.service file to /etc/systemd/system
+		// 4. run systemctl daemon-reload
+
+		return nil
+	},
+}
+
+var uninstallCmd = &cobra.Command{
+	Use:               "uninstall",
+	Short:             "Uninstall the startup process for Cloudfuse. Requires running as admin.",
+	Long:              "Uninstall the startup process for Cloudfuse. Requires running as admin.",
+	SuggestFor:        []string{"uninst", "uninstal"},
+	Example:           "cloudfuse service uninstall",
+	FlagErrorHandling: cobra.ExitOnError,
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		// 1. find and remove cloudfuse.service from /etc/systemd/system and run systemctl daemon-reload
+		// 2. handle errors
+
+		return nil
+	},
+}
+
+//--------------- command section ends
+
+func init() {
+	rootCmd.AddCommand(serviceCmd)
+	serviceCmd.AddCommand(installCmd)
+	serviceCmd.AddCommand(uninstallCmd)
+}
--- a/cmd/service_linux.go
+++ b/cmd/service_linux.go
@@ -66,39 +66,25 @@ var installCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) error {
 
 		// 1. get the cloudfuse.service file from the setup folder and collect relevant data (user, mount, config)
-		serviceFile, err := os.Open("./setup/cloudfuse.service")
 
+		serviceData, err := collectServiceData("./setup/cloudfuse.service")
 		if err != nil {
-			fmt.Println("Error opening file:", err)
-			return err
+			return fmt.Errorf("error collecting data from cloudfuse.service file due to the following error: ", err.Error())
 		}
 
-		defer serviceFile.Close()
-
-		scanner := bufio.NewScanner(serviceFile)
-		serviceData := make(map[string]string)
-		for scanner.Scan() {
-			line := scanner.Text()
-			if strings.Contains(line, "Environment=") {
-				parts := strings.SplitN(line, "=", 3)
-				key := strings.TrimSpace(parts[1])
-				value := strings.TrimSpace(parts[2])
-				serviceData[key] = value
-			}
-			if strings.Contains(line, "User=") {
-				parts := strings.SplitN(line, "=", 2)
-				key := strings.TrimSpace(parts[0])
-				value := strings.TrimSpace(parts[1])
-				serviceData[key] = value
+		//TODO: set a default mount and config path when stubbed examples are found in the file.
+		mountPath := serviceData["MoutingPoint"]
+		if mountPath == "/path/to/mounting/point" {
+			fmt.Printf("creating mount folder in $HOME/cloudfuseMount")
+			userAddCmd := exec.Command("mkdir", "$HOME/cloudfuseMount")
+			err := userAddCmd.Run()
+			if err != nil {
+				return fmt.Errorf("failed to create default mount folder due to following error: [%s]", err.Error())
 			}
 		}
-		if err := scanner.Err(); err != nil {
-			fmt.Println("Error reading file:", err)
-			return err
-		}
 
-		// 2. retrieve the user account from cloudfuse.service file and make it if it doesn't exist
-		value := serviceData["User"]
+		// 2. retrieve the user account from cloudfuse.service file and create it if it doesn't exist
+		user := serviceData["User"]
 		usersList, err := os.Open("/etc/passwd")
 		if err != nil {
 			return fmt.Errorf("failed to open /etc/passwd due to following error: [%s]", err.Error())
@@ -109,13 +95,13 @@ var installCmd = &cobra.Command{
 
 		for scanner.Scan() {
 			line := scanner.Text()
-			if strings.HasPrefix(line, value) {
+			if strings.HasPrefix(line, user) {
 				foundUser = true
 			}
 		}
 		if !foundUser {
 			//create the user
-			userAddCmd := exec.Command("useradd", "-m", value)
+			userAddCmd := exec.Command("useradd", "-m", user)
 			err := userAddCmd.Run()
 			if err != nil {
 				return fmt.Errorf("failed to create user due to following error: [%s]", err.Error())
@@ -158,6 +144,42 @@ var uninstallCmd = &cobra.Command{
 }
 
 //--------------- command section ends
+
+func collectServiceData(serviceFilePath string) (map[string]string, error) {
+	serviceFile, err := os.Open("./setup/cloudfuse.service")
+
+	if err != nil {
+		fmt.Println("Error opening file:", err)
+		return nil, err
+	}
+
+	defer serviceFile.Close()
+
+	scanner := bufio.NewScanner(serviceFile)
+	serviceData := make(map[string]string)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.Contains(line, "Environment=") {
+			parts := strings.SplitN(line, "=", 3)
+			key := strings.TrimSpace(parts[1])
+			value := strings.TrimSpace(parts[2])
+			serviceData[key] = value
+		}
+		if strings.Contains(line, "User=") {
+			parts := strings.SplitN(line, "=", 2)
+			key := strings.TrimSpace(parts[0])
+			value := strings.TrimSpace(parts[1])
+			serviceData[key] = value
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		fmt.Println("Error reading file:", err)
+		return nil, err
+	}
+	return serviceData, nil
+}
+
+//TODO: add wrapper function for collecting data, creating user, setting default paths, running commands.
 
 func init() {
 	rootCmd.AddCommand(serviceCmd)
--- a/cmd/service_linux.go
+++ b/cmd/service_linux.go
@@ -31,6 +31,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"os/user"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


cmd/service_linux.go
```diff
@@ -0,0 +1,340 @@
...
+		dir, err := os.Getwd()
+		if err != nil {
+			return fmt.Errorf("error: [%s]", err.Error())
+		}
+
+		// assumes dir is in cloudfuse repo dir
+		serviceData, err := collectServiceData(fmt.Sprintf("%s/setup/cloudfuse.service", dir))
+		if err != nil {
+			return fmt.Errorf("error collecting data from cloudfuse.service file due to the following error: [%s]", err)
+		}
```
foodprocessor, line 83: For deployment, we need to add the `cloudfuse.service` file to the list of deployed files (in `.goreleaser.yaml`), and something like this should work. We may need to use `filepath.Dir(os.Executable())` ([documented here](https://pkg.go.dev/os@master#Executable)), instead of `os.Getwd()` to reliably get the path we need.

cmd/service_linux.go
```diff
@@ -0,0 +1,340 @@
...
+		mountExists := common.DirectoryExists(mountPath)
+		if !mountExists {
+			return fmt.Errorf("error, the mount path provided does not exist")
+			// TODO: add useage output upon failure with input
+		}
+		isDirEmpty := common.IsDirectoryEmpty(mountPath)
+		if !isDirEmpty {
+			return fmt.Errorf("error, the mount path provided is not empty")
+			// TODO: add useage output upon failure with input
+		}
```
foodprocessor, line 98: I don't think we need to check this.

cmd/service_linux.go
```diff
@@ -0,0 +1,340 @@
...
+
+		configExists := common.DirectoryExists(configPath)
```
foodprocessor, line 100: Although this works, I don't love that `DirectoryExists` is checking existence for a _file_.
I found another place where we check if the config file exists, and we use os.Stat directly:
```
if _, err := os.Stat(options.ConfigFile); errors.Is(err, fs.ErrNotExist) {
	return errors.New("config file does not exist")
}
```

cmd/service_linux.go
```diff
@@ -0,0 +1,340 @@
...
+		systemctlEnableCmd := exec.Command("sudo", "systemctl", "enable", "cloudfuse.service")
+		err = systemctlEnableCmd.Run()
+		if err != nil {
+			return fmt.Errorf("failed to run 'systemctl daemon-reload' command due to following error: [%s]", err.Error())
+		}
+		return nil
+	},
+}
+
+var uninstallCmd = &cobra.Command{
```
foodprocessor, line 144: The uninstall command will have to require the mount path to know which mount's startup service we need to remove.

cmd/service_linux.go
```diff
@@ -0,0 +1,340 @@
...
+
+		err = modifySericeFile(mountPath, dir)
+		if err != nil {
+			return fmt.Errorf("error when attempting to write defaults into service file: [%s]", err.Error())
+		}
+
+		err = modifySericeFile(configPath, dir)
+		if err != nil {
+			return fmt.Errorf("error when attempting to write defaults into service file: [%s]", err.Error())
+		}
```
foodprocessor, line 113: Since we're creating a separate service file for each mount that the user wants to mount on startup, we'll want to

- not change our template cloudfuse.service file (we should write the modified unit file data straight to the destination)
- name each new service/unit file after the mount path, so they don't conflict, and so we can find them to remove them when uninstalling

cmd/service_linux.go
```diff
@@ -0,0 +1,340 @@
...
+
+		// 2. retrieve the newUser account from cloudfuse.service file and create it if it doesn't exist
+
+		serviceUser := serviceData["User"]
+		setUser(serviceUser)
```
foodprocessor, line 118: On Linux I don't think we need a dedicated user.

cmd/service_linux.go
```diff
@@ -0,0 +1,340 @@
...
+			return fmt.Errorf("failed to run 'systemctl daemon-reload' command due to following error: [%s]", err.Error())
+		}
+
+		return nil
+	},
+}
+
+//--------------- command section ends
+
+func collectServiceData(serviceFilePath string) (map[string]string, error) {
```
foodprocessor, line 172: I don't know, but I feel like these helper functions could be a lot shorter, and more readable. I feel like these details should be abstracted away by a standard library function that does most or all of this for us...

foodprocessor (changes requested):

cmd/service_linux.go
Dabnsky: cloudfuse outputs and error and will not mount if the provided mount path is not empty

cmd/service_linux.go
Dabnsky: While failing to specify a user with the appropriate permissions doesn't stop the cloudfuse service from running and working, it opens up a large security risk involving the default permissions that are used when the service is run.

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+func newSericeFile(mountPath string, configPath string, dir string) (string, error) {
+
+	//mountpath or config?
+	var oldString string
+	var newString string
+
+	oldConfigStr := "Environment=ConfigFile=/path/to/config/file/config.yaml"
+	newConfigStr := fmt.Sprintf("Environment=ConfigFile=%s", common.JoinUnixFilepath(dir, configPath))
+
+	oldMountStr := "Environment=MoutingPoint=/path/to/mounting/point"
```
jfantinhardesty, line 169: This and a few others should be MountingPoint.

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		fmt.Println("Error reading file:", err)
+		return nil, err
+	}
+	return serviceData, nil
+}
+
+func newSericeFile(mountPath string, configPath string, dir string) (string, error) {
```
jfantinhardesty: This should be newServiceFile

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+	var newString string
+
+	oldConfigStr := "Environment=ConfigFile=/path/to/config/file/config.yaml"
+	newConfigStr := fmt.Sprintf("Environment=ConfigFile=%s", common.JoinUnixFilepath(dir, configPath))
+
+	oldMountStr := "Environment=MoutingPoint=/path/to/mounting/point"
+	newMountStr := fmt.Sprintf("Environment=MoutingPoint=%s", common.JoinUnixFilepath(dir, mountPath))
+
+	// open service defaultFile for read write
+	defaultFile, err := os.OpenFile("./setup/cloudfuse.service", os.O_RDWR, 0644)
```
jfantinhardesty, line 173: Could you set this as a constant to refer to at the top of the file.

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+		// Append the (possibly modified) line to the slice
+		lines = append(lines, line)
+	}
+	// Check for errors during file reading
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("error reading file: [%s]", err.Error())
+	}
+
+	folderList := strings.Split(mountPath, "/")
+	newFile, err := os.Create("/etc/systemd/system/" + folderList[len(folderList)-1] + ".service")
```
jfantinhardesty: Could you set this as a constant to refer to at the top of the file.

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/Seagate/cloudfuse/common"
+	"github.com/spf13/cobra"
+)
+
+type serviceOptions struct {
+	ConfigFile string
+	MountPath  string
+}
+
+var servOpts serviceOptions
```
jfantinhardesty, line 47: This variable is never used.

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+	SuggestFor:        []string{"ser", "serv"},
+	Example:           "cloudfuse service install",
+	FlagErrorHandling: cobra.ExitOnError,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return errors.New("missing command options\n\nDid you mean this?\n\tcloudfuse service install\n\nRun 'cloudfuse service --help' for usage")
+	},
+}
+
+var installCmd = &cobra.Command{
+	Use:               "install [mount path] [config path]",
```
jfantinhardesty, line 63: Could you change this to use flags like mount-path and config-path instead of relying on order?

cmd/service_linux.go
jfantinhardesty: In this function you assume that configPath is just located in the current directory dir. Could you instead get the full path to the config file, since the user needs to pass the location of configPath anyway.

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+		if strings.Contains(line, "MoutingPoint") {
+			// Modify the line by replacing the old string with the new string
+			line = strings.ReplaceAll(line, oldMountStr, newMountStr)
+		}
+		if strings.Contains(line, "ConfigFile") {
+			line = strings.ReplaceAll(line, oldConfigStr, newConfigStr)
+		}
+
+		// add the -o default_permissions if not present
+		if strings.Contains(line, "ExecStart") && !strings.Contains(line, "-o allow_other") {
```
jfantinhardesty, line 196: Why are you checking for this if we write the config file. Shouldn't the written config file already have allow_other?

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+
+		serviceFile, err := newSericeFile(mountPath, configPath, dir)
+		if err != nil {
+			return fmt.Errorf("error when attempting to create service file: [%s]", err.Error())
+		}
+
+		// 2. retrieve the newUser account from cloudfuse.service file and create it if it doesn't exist
+
+		// assumes dir is in cloudfuse repo dir
+		serviceData, err := collectServiceData(fmt.Sprintf("%s/setup/cloudfuse.service", dir))
```
jfantinhardesty, line 108: You assume that this directory exists. But if the user just runs the installer for cloudfuse this file will not exist, you will need to write it for them.

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+	if err != nil {
+		return "", fmt.Errorf("error opening file: [%s]", err.Error())
+	}
+	defer defaultFile.Close()
+
+	scanner := bufio.NewScanner(defaultFile)
+	var lines []string
+
+	// Read the file line by line
+	for scanner.Scan() {
```
jfantinhardesty, line 183: Rather than doing string replacement, why not just have a template for the service file you would like to write, and then add the configFile and mountPath information to that template and then write the template file? Possibly using text/template or something similar.

jfantinhardesty (changes requested):

cmd/service_linux.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+
+		// use the current user for reference on permissions
+		usr, err := user.Current()
+		if err != nil {
+			return fmt.Errorf("error getting looking up current user: [%s]", err.Error())
+		}
+
+		// get a list of group from reference user
+		groups, err := usr.GroupIds()
```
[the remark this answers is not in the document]
Dabnsky: the user at least needs the current user added to its groups to be able to access and use mount and config file typically located at the end user's home directory. There doesn't seem to be a way around using the usermod command for this.

cmd/service_linux.go
Dabnsky: I am using the setup/cloudfuse.service file as a template. you mentioned in another comment about how this file may not exist if the installer was used. How would you suggest I have a template reference in that scenario? Should I have the template hard coded and reference it as needed?

cmd/service_linux.go
Dabnsky: Are you suggesting I should have a hard coded template to reference? or is there another way to have a template I can always reference?

jfantinhardesty: Yeah, you could just use a template either in the form of a string that you add the mountpoint and config too or something like text/template. Possibly something like this https://github.com/pioz/god.

cmd/service_linux.go
jfantinhardesty: Yeah I would not rely on that file existing. The executable should be able to write the file to systemd. So like suggested above I would have that hard coded in a string or template and then write the file.

cmd/service_linux.go
Dabnsky: There is no guarantee that the config file will have this option. It's easier to supply this option here instead of checking for it in the config file.

jfantinhardesty: I guess what I mean is if we are writing the systemd file ourselves, then we don't need to read it and then determine if we should add it if we are always adding that to the systemd file.

cmd/service_linux.go
```diff
@@ -0,0 +1,335 @@
...
+	if err != nil {
+		fmt.Errorf("error creating new service file: [%s]", err.Error())
+	}
+
+	folderList := strings.Split(mountPath, "/")
+	serviceName := folderList[len(folderList)-1] + ".service"
+	newFile, err := os.Create("/etc/systemd/system/" + serviceName)
+	if err != nil {
+		return "", fmt.Errorf("error creating new service file: [%s]", err.Error())
+	}
```
foodprocessor, line 216: Could you overwrite it, in case they want to use the install command to update an existing service?
Or would you rather they uninstall first?

foodprocessor (changes requested):
Great improvements!!

cmd/service_linux.go
foodprocessor: Sorry, you're right. I forgot that on Linux, adding a user is exactly how this is done.

cmd/service_linux.go
```diff
@@ -0,0 +1,335 @@
...
+	mountPathGroup, err := user.LookupGroupId(fmt.Sprint(mountGroupID))
+	if err != nil {
+		return fmt.Errorf("failed to lookup group: %v", err)
+	}
+
+	_, err = user.Lookup(serviceUser)
+	if err != nil {
+		if strings.Contains(err.Error(), "unknown user") {
+			//create the user
+			userAddCmd := exec.Command("sudo", "useradd", "-m", serviceUser)
```
jfantinhardesty, line 268: I think you should user "-r" instead of "-m" since "-m" will create a user that has a home directory and we probably don't want to do that.

cmd/service_linux.go
```diff
@@ -0,0 +1,335 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to create user due to following error: [%s]", err.Error())
+			}
+
+			//set set folder permission on the mount path
+			chmodCmd := exec.Command("sudo", "chmod", "770", mountPath)
+			err = chmodCmd.Run()
+			if err != nil {
+				return fmt.Errorf("failed set permisions on mount path due to following error: [%s]", err.Error())
+			}
```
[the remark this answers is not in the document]
jfantinhardesty: I agree. I think that makes it a bit easier and prevents us from doing something wrong to the users system.

cmd/service_linux.go
```diff
@@ -0,0 +1,335 @@
...
+			return fmt.Errorf("failed to run 'systemctl daemon-reload' command due to following error: [%s]", err.Error())
+		}
+		return nil
+	},
+}
+
+//--------------- command section ends
+
+func newServiceFile(mountPath string, configPath string, serviceUser string) (string, error) {
+	serviceTemplate := ` [Unit]
```
jfantinhardesty, line 178: There shouldn't be a space here before [Unit]

cmd/service_linux.go
```diff
@@ -0,0 +1,335 @@
...
+	Description=Cloudfuse is an open source project developed to provide a virtual filesystem backed by S3 or Azure storage.
+	After=network-online.target
+	Requires=network-online.target
+
+	[Service]
+	# User service will run as.
+	User={{.ServiceUser}}
+	# Path to the location Cloudfuse will mount to. Note this folder must currently exist.
+	Environment=MountingPoint={{.MountPath}}
```
jfantinhardesty, line 187: You could simplify this environment section. I only did it this way to make it easy for a user to edit. But if you write it then you could just put the mountPath in the mount and unmount commands directly.

cmd/service_linux.go
```diff
@@ -0,0 +1,335 @@
...
+			if err != nil {
+				return fmt.Errorf("failed set permisions on mount path due to following error: [%s]", err.Error())
+			}
+
+		} else {
+			fmt.Printf("An error occurred: %v\n", err)
+		}
+	} else {
+		//add group to serviceUser group
+		usermodCmd := exec.Command("sudo", "usermod", "-aG", configFileGroup.Name, serviceUser)
```
jfantinhardesty, line 298: If the user exists, we probably should not modify the group permissions and assume the user has them set in a way to run cloudfuse. And if there is an error we should tell them how to fix that error.

jfantinhardesty (changes requested):
Left a few more comments to simplify some of this. Also, I tend to agree with Michael that we probably should not run a lot of these commands ourselves, and instead suggest them to the user. Also, there are a variety of lint and spelling errors that should be fixed.

cmd/service_linux.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		isDirEmpty := common.IsDirectoryEmpty(mountPath)
+		if !isDirEmpty {
+			return fmt.Errorf("error, the mount path provided is not empty")
+			// TODO: add useage output upon failure with input
+		}
+
+		_, err = os.Stat(configPath)
+		if errors.Is(err, fs.ErrNotExist) {
+			return fmt.Errorf("error, the configfile path provided does not exist") // TODO: add useage output upon failure with input
```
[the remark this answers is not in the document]
Dabnsky: yes it is

cmd/service_linux.go
Dabnsky: yes, currently the service file gets overwritten if a redundant mount path is provided. I think that's fine

Dabnsky: yes, currently the service file gets overwritten if a redundant mount path is provided. I think that's fine

Dabnsky: The install command will now delete and create the service file if it already exists allowing an overwrite.

cmd/service_linux.go
Dabnsky: it looks like cloudfuse automatically tries to create the home folder of the service user if it doesn't exist, which I don't fully understand. but starting the service is failing with the service user when it doesn't have a home directory. It runs into permissions issues upon attempting to do a mkdir command to set it up

cmd/service_linux.go
```diff
@@ -0,0 +1,273 @@
...
+	var absPath string
+	var err error
+	if !filepath.IsAbs(leaf) {
+		absPath, err = filepath.Abs(leaf)
+		if err != nil {
+			return "", fmt.Errorf("couldn't format the path string", err.Error())
+		}
+	}
+	return absPath, err
+}
```
foodprocessor, line 259: Is there any downside to just calling `filepathAbs` without checking `filepath.IsAbs`?
If not, maybe we can drop this function altogether and just call `filepath.Abs` directly.

cmd/service_linux.go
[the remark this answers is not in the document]
foodprocessor: There's still one `TODO` comment in the code mentioning usage prints. Is that one also okay to delete?

cmd/service_linux.go
```diff
@@ -0,0 +1,335 @@
...
+			mountPath = common.JoinUnixFilepath(dir, mountPath)
+		}
+
+		if strings.Contains(configPath, ".") || strings.Contains(configPath, "..") {
+			configPath = common.JoinUnixFilepath(dir, configPath)
+		}
+
+		mountExists := common.DirectoryExists(mountPath)
+		if !mountExists {
+			return fmt.Errorf("error, the mount path provided does not exist")
```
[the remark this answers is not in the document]
foodprocessor: There are still multiple error returns that could use this change.

cmd/service_linux.go
```diff
@@ -0,0 +1,273 @@
...
+var mountPath string
+var configPath string
+var serviceUser string
+
+var installCmd = &cobra.Command{
+	Use:               "install",
+	Short:             "Installs a service file for a single mount with Cloudfuse. Requires elevated permissions.",
+	Long:              "Installs a service file for a single mount with Cloudfuse. Requires elevated permissions.",
+	SuggestFor:        []string{"ins", "inst"},
+	Example:           "cloudfuse service install --mount-path=<path/to/mount/point> --config-file=<path/to/config/file> --user=<username>",
```
foodprocessor, line 71: Could the user default to `cloudfuse`, and the switch become optional?

cmd/service_linux.go
```diff
@@ -0,0 +1,273 @@
...
+		ServiceUser: serviceUser,
+	}
+
+	tmpl, err := template.New("service").Parse(serviceTemplate)
+	if err != nil {
+		return "", fmt.Errorf("error creating new service file: [%s]", err.Error())
+	}
+
+	folderList := strings.Split(mountPath, "/")
+	serviceName := "cloudfuse-" + folderList[len(folderList)-1] + ".service"
```
foodprocessor, line 200: Could you use the whole path (every element of `folderList`) here, to truly guarantee a unique name?

cmd/service_linux.go
```diff
@@ -0,0 +1,335 @@
...
+		ServiceUser: serviceUser,
+	}
+
+	tmpl, err := template.New("service").Parse(serviceTemplate)
+	if err != nil {
+		fmt.Errorf("error creating new service file: [%s]", err.Error())
+	}
+
+	folderList := strings.Split(mountPath, "/")
+	serviceName := folderList[len(folderList)-1] + ".service"
```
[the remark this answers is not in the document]
foodprocessor: The newest code here is only using the final element of the path. Could you use the whole path, to truly guarantee a unique name?

cmd/service_linux.go
```diff
@@ -0,0 +1,273 @@
...
+		delServFileCmd := exec.Command("rm", servicePath)
+		err = delServFileCmd.Run()
+		if err != nil {
+			return "", fmt.Errorf("failed to replace the service file due to the following error: [%s]", err.Error())
+		}
+		newFile, err = os.Create(servicePath)
+		if err != nil {
+			return "", fmt.Errorf("error creating new service file: [%s]", err.Error())
+		}
+	}
```
foodprocessor, line 219: Could you simplify this logic by only calling os.Create once?

foodprocessor (changes requested):
Looks great!
Almost done.

cmd/service_linux.go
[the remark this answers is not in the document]
foodprocessor: There are still many error messages that start with the word "error." Since these messages are prepended with "Error:", I think "failed to ..." is a better wording for those cases one failure is caused by another.

cmd/service_linux.go
```diff
@@ -0,0 +1,263 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		// get absolute path of provided relative mount path
+		var err error
+		mountPath, err = getAbsPath(serviceName)
+		if err != nil {
+			return err
+		}
+		serviceName := strings.Replace(mountPath, "/", "-", -1)
+		serviceName = strings.TrimPrefix(serviceName, "-")
```
foodprocessor, line 141: We probably want the service name to include "cloudfuse". Could you prepend that to it?

cmd/service_linux.go
```diff
@@ -0,0 +1,263 @@
...
+	Short:             "Uninstall a startup process for Cloudfuse.",
+	Long:              "Uninstall a startup process for Cloudfuse.",
+	SuggestFor:        []string{"uninst", "uninstal"},
+	Example:           "cloudfuse service uninstall --mount-path=<path/to/mount/path>",
+	FlagErrorHandling: cobra.ExitOnError,
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		// get absolute path of provided relative mount path
+		var err error
+		mountPath, err = getAbsPath(serviceName)
```
foodprocessor, line 136: I think the intention here was to call getAbsPath on `mountPath` instead of `serviceName`.

cmd/service_linux.go
```diff
@@ -0,0 +1,263 @@
...
+
+	tmpl, err := template.New("service").Parse(serviceTemplate)
+	if err != nil {
+		return "", fmt.Errorf("error creating new service file: [%s]", err.Error())
+	}
+
+	serviceName := strings.Replace(mountPath, "/", "-", -1)
+	serviceName = strings.TrimPrefix(serviceName, "-")
+	serviceFile := serviceName + ".service"
+	serviceFilePath := "/etc/systemd/system/" + serviceFile
```
foodprocessor, line 200: This code is duplicated. Do we need some constants and a function?

cmd/service_linux.go
```diff
@@ -0,0 +1,263 @@
...
+
+	err = os.Remove(serviceFilePath)
+	if err != nil && !os.IsNotExist(err) {
+		return "", fmt.Errorf("failed to replace the service file due to the following error: [%s]", err.Error())
+	}
```
foodprocessor, line 205: Won't this return an error when the service file doesn't already exist?

foodprocessor (changes requested):

cmd/service_linux.go
Dabnsky: This error should take place only when the error is anything other than isNotExist. That is my interpretation of this condition. is that not a correct evaluation?

jfantinhardesty (changes requested):
Looks pretty close. A few things to fix:

1. The linter is giving errors. Fix the lint issues in this file so that all of the tests in GitHub actions are green.
2. The uninstall command does not work. You set serviceName in the uninstall command options when the user passes in mount-path but then don't use it and refer to the mountPath variable in the uninstallCmd function so then you are not able to uninstall the service. You should probably remove the serviceName and just use the mountPath variable. Test to make sure that this change works.
3. On line 220, that fmt.Println should not end with a newline character since this is newline it will insert one automatically.
4. You should merge main into this branch and then update the copyright header so it is 2023-2025.

cmd/service_linux.go
```diff
@@ -0,0 +1,252 @@
...
+		systemctlEnableCmd := exec.Command("systemctl", "enable", serviceName)
+		err = systemctlEnableCmd.Run()
+		if err != nil {
+			return fmt.Errorf("failed to run 'systemctl daemon-reload' command due to following error: [%s]", err.Error())
+		}
+		return nil
+	},
+}
+
+var serviceName string
```
foodprocessor, line 119: This variable declaration looks like it's not used. Can you drop it?

cmd/service_linux.go
foodprocessor: You're right. My comment is outdated. It was written about the code before [this change](https://github.com/Seagate/cloudfuse/commit/1ddeb0aa84fbdbd564772a4db97cf87296b376ec).

foodprocessor (approved):
Looks great!

cmd/service_linux.go
Dabnsky: helper function removed

jfantinhardesty (approved):

cmd/service_linux.go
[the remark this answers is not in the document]
Dabnsky: replaced the rest of them. let me know if I missed one.

cmd/service_linux.go
```diff
@@ -0,0 +1,246 @@
...
+	Long:              "Uninstall a startup process for Cloudfuse.",
+	SuggestFor:        []string{"uninst", "uninstal"},
+	Example:           "cloudfuse service uninstall --mount-path=<path/to/mount/path>",
+	FlagErrorHandling: cobra.ExitOnError,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// get absolute path of provided relative mount path
+
+		mountPath, err := filepath.Abs(mountPath)
+		if err != nil {
+			return fmt.Errorf("couldn't couldn't determine absolute path from string [%s]", err.Error())
```
foodprocessor, line 131: typ-o

Dabnsky:
datalake error fixed in separate PR that was holding up the last code coverage test

PR #358, opened 2024-11-02
Use Bucket Size for Filesystem Stats

foodprocessor:

- Add flag to allow users to choose whether `StatFs` will be populated by file cache or cloud storage.
- Provide `StatFs` storage usage statistics from cloud storage by default (should we flip this, or should the flag be in s3storage?)
- Correct `StatFs` math in file cache.
- Use same `StatFs` function in file cache regardless of platform (and use `getAvailableSize()` for platform-specific code).
- Change libfuse `StatFs` response to handle both file cache and cloud storage data.
- Change s3 call to `HeadBucket` to capture response and parse size (works for Lyve Cloud only).
- Create new function `bucketExists()` to replace old `headBucket` wrapper.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [x] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #


setup/baseConfig.yaml
```diff
@@ -102,6 +102,7 @@ file_cache:
   max-size-mb: <maximum cache size allowed. Default - 80% of free disk space>
   high-threshold: <% disk space consumed which triggers eviction. This parameter overrides 'timeout-sec' parameter and cached files will be removed even if they have not expired. Default - 80>
   low-threshold: <% disk space consumed which triggers eviction to stop when previously triggered by the high-threshold. Default - 60>
+  usage-from-cache: true|false <use file cache to report drive size and storage statistics (StatFs)>
```
jfantinhardesty, line 105: Should we clarify that this should only be false when using Seagate Lyve Cloud?

jfantinhardesty (commented):
Great fix for the size of the bucket. On question, is there a performance concern with how often the OS may call statfs. Seems like it might be a bit slow to respond each time and was wondering if this has caused a problem with Network Optix or if you see a problem with the additional latency?

setup/baseConfig.yaml
foodprocessor: Yeah, I can add that note.
We can also discuss the various options for this flag (default the other way? Or move the flag to the s3storage component?)
I'm starting to think it might need to move to s3storage...

foodprocessor:
> Great fix for the size of the bucket. On question, is there a performance concern with how often the OS may call statfs. Seems like it might be a bit slow to respond each time and was wondering if this has caused a problem with Network Optix or if you see a problem with the additional latency? Also it appears that the test TestStatFs is failing on both Linux and Windows.

I have only tested this on Windows. I have not seen performance issues, and the OS seems to only call StatFs at the expected times (when viewing the mount properties and when an application calls StatFs explicitly).

jfantinhardesty (approved):
Looks good! I like the config option in s3storage.

PR #370, opened 2024-11-18
Fix open append

### LINKED ISSUE #369: Unable to append to file
### Problem

When opening a file with O_APPEND, writes are not able to be appended to the file and 

### Expected behavior

Data should be able to be appended to files.

### How to reproduce

Run the following command in a cloudfuse mounted directory with a file test.txt already existing. `echo "Appending text" >> ./test.txt`. This will then give the error `bash: echo: write error: Input/output error`

### Logs

_No response_

### Deployment information

Latest cloudfuse version 1.7.1 using file_cache config.

### Additional information

_No response_

jfantinhardesty:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This fixes a bug where cloudfuse was unable to append to files opened with the O_APPEND flag.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Closes #369

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -843,7 +843,7 @@ func (fc *FileCache) downloadFile(handle *handlemap.Handle) error {
 		}
 
 		// Open the file in write mode.
-		f, err = common.OpenFile(localPath, os.O_CREATE|os.O_RDWR, fMode)
+		f, err = common.OpenFile(localPath, flags, fMode)
 		if err != nil {
 			log.Err("FileCache::downloadFile : error creating new file %s [%s]", handle.Path, err.Error())
 			return err
@@ -966,6 +966,9 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
 	handle := handlemap.NewHandle(options.Name)
 	handle.SetValue("openFileOptions", openFileOptions{flags: options.Flags, fMode: options.Mode})
 	// Increment the handle count in this lock item as there is one handle open for this now
+	if options.Flags&os.O_APPEND != 0 {
+		handle.Flags.Set(handlemap.HandleOpenedAppend)
+	}
 	flock.Inc()
 
 	return handle, nil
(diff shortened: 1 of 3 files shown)


component/file_cache/file_cache.go
```diff
@@ -966,6 +966,9 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
 	handle := handlemap.NewHandle(options.Name)
 	handle.SetValue("openFileOptions", openFileOptions{flags: options.Flags, fMode: options.Mode})
 	// Increment the handle count in this lock item as there is one handle open for this now
+	if options.Flags&os.O_APPEND != 0 {
+		handle.Flags.Set(handlemap.HandleOpenedAppend)
+	}
```
foodprocessor, line 971: This looks good, but you split a comment (above) from its code (below).

component/file_cache/file_cache.go
```diff
@@ -843,7 +843,7 @@ func (fc *FileCache) downloadFile(handle *handlemap.Handle) error {
 		}
 
 		// Open the file in write mode.
-		f, err = common.OpenFile(localPath, os.O_CREATE|os.O_RDWR, fMode)
+		f, err = common.OpenFile(localPath, flags, fMode)
```
foodprocessor, line 846: The flags we use here are for downloading object data from the cloud. We should not use append for that, right?

foodprocessor: Maybe a "case 1" unit test would have caught this. Do you want to add one?

foodprocessor (changes requested):

component/file_cache/file_cache.go
jfantinhardesty: Fixed.

component/file_cache/file_cache.go
jfantinhardesty: Good catch. Actually, this was not needed to fix O_APPEND and accidently included while I was trying to track the permissions issue down. So I have reverted this change.

foodprocessor (approved):
Thanks!

PR #404, opened 2025-01-03
Update open file handles in RenameFile

foodprocessor:
* Update open file handles in RenameFile
* Add failure recovery for partial renames to prevent data loss
* Add tests

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -647,7 +647,11 @@ func (fc *FileCache) RenameDir(options internal.RenameDirOptions) error {
 					sflock.Lock()
 				}
 				// complete local rename
-				fc.renameCachedFile(path, newPath, sflock, dflock)
+				err := fc.renameCachedFile(path, newPath, sflock, dflock)
+				if err != nil {
+					// there's really not much we can do to handle the error, so just log it
+					log.Err("FileCache::RenameDir : %s file rename failed. Directory state is inconsistent!", path)
+				}
 				sflock.Unlock()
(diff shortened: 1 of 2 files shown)


component/file_cache/file_cache.go
```diff
@@ -1366,16 +1389,22 @@ func (fc *FileCache) GetAttr(options internal.GetAttrOptions) (*internal.ObjAttr
...
 
+	// acquire file locks
 	sflock := fc.fileLocks.Get(options.Src)
-	sflock.Lock()
-	defer sflock.Unlock()
-
 	dflock := fc.fileLocks.Get(options.Dst)
-	dflock.Lock()
+	// prevent deadlock
+	if options.Src < options.Dst {
```
Ferelith-maker, line 1400: Can you explain how src > dst works?

foodprocessor: It's just checking which string comes first, in alphabetical / lexical order.
The point of doing this is to make sure that we always lock the semaphores in the same order, to avoid deadlock (where one thread locks A and waits to lock B while another thread locks B and waits to lock A).

Ferelith-maker (approved):
Looks good to me

jfantinhardesty (approved):

PR #419, opened 2025-01-16
File cache test filenames don't need to be unique anymore!

foodprocessor:
## ✅ What

* A bit of code cleanup.
* Simplify code that clears the lazy flag when a file handle is closed.
* Only close test file handles when it's needed (for Windows). 
* With this change, test cleanup runs correctly, and so tests can reuse filenames without fear of collisions.
* Add assert to make unit tests fail when cleanup fails
 
## 🤔 Why
 
In one case handles need to be closed for tests to run correclty, but in most cases it's needed so test cleanup can delete the test files.
 
## 👩‍🔬 How to validate if applicable
 
Run file cache unit tests on Windows and Linux, and see no errors complaining about failure to cleanup the test.


jfantinhardesty (approved):
Love cleaning up our code!

PR #423, opened 2025-01-23
Size tracking component

jfantinhardesty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This adds a new size tracker component which tracks the size of the mounted directory to prevent calls to the cloud to get the size of the mount. This in particular is useful when mounting subdirectories as there is no call in S3 to retrieve the size of the storage. This component goes in between the file_cache and attr_cache to track write information.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #


foodprocessor (approved):
That's quite the testing infrastructure! Good work!

component/size_tracker/size_tracker.go
```diff
@@ -0,0 +1,388 @@
...
+		log.Err("SizeTracker::WriteFile : Unable to get attr for file %s. Current tracked size is invalid. Error: : %v", options.Handle.Path, getAttrErr1)
+	}
+
+	bytesWritten, err := st.NextComponent().WriteFile(options)
+	if err != nil {
+		return bytesWritten, err
+	}
+
+	var newSize int64
+	attr, getAttrErr2 := st.NextComponent().GetAttr(internal.GetAttrOptions{Name: options.Handle.Path})
```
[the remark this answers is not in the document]
jfantinhardesty: Yeah, good idea to avoid an extra call. I will make that change.

component/size_tracker/size_tracker.go
```diff
@@ -0,0 +1,377 @@
...
+
+	return err
+}
+
+func (st *SizeTracker) Priority() internal.ComponentPriority {
+	return internal.EComponentPriority.LevelOne()
+}
+
+// OnConfigChange : If component has registered, on config file change this method is called
+func (st *SizeTracker) OnConfigChange() {
```
Dabnsky, line 118: This function being empty is interesting and notable. Can you explain how this is being used?

component/size_tracker/size_tracker.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func (st *SizeTracker) RenameDir(options internal.RenameDirOptions) error {
+	// Rename dir should not allow renaming files into a directory that already exists so we should not
+	// need to update the size here.
+	return st.NextComponent().RenameDir(options)
+}
+
+// File operations
+func (st *SizeTracker) CreateFile(options internal.CreateFileOptions) (*handlemap.Handle, error) {
```
Dabnsky, line 128: What kind of file would be created here? is this different from the file that gets created in the CreateSizeJournal file from journal.go?

component/size_tracker/size_tracker.go
```diff
@@ -0,0 +1,377 @@
...
+	if getAttrErr == nil {
+		origSize = attr.Size
+	}
+
+	err := st.NextComponent().TruncateFile(options)
+	newSize := options.Size - origSize
+
+	var journalErr error
+	// File already exists and truncate succeeded subtract difference in file size
+	if err == nil && getAttrErr == nil && newSize < 0 {
```
Dabnsky, line 219: This is a very cool way to do your error checking. If there are any errors, they are returned at the end and you don't have to catch and output them on the spot. I think I like this.

component/size_tracker/size_tracker.go
jfantinhardesty: When you create a new component, some of these functions are generated automatically. This being one such function. So this would be used if when the user updates their config file if this should change something. But it currently doesn't do anything in this component.

test/s3cleanup/s3cleanup_test.go
```diff
@@ -0,0 +1,115 @@
...
+			})
+		}
+
+		if len(objectsToDelete) > 0 {
+			err = deleteObjects(ctx, client, bucket, objectsToDelete)
+			if err != nil {
+				log.Fatal(err)
+			}
+		}
+		objectsToDelete = nil
```
Dabnsky, line 99: do we want to do another touch on the cloud to assert that it is now empty?

test/e2e_tests/statfs_linux.go
Dabnsky, line 1: how does doing a statfs relate to the size tracking of the file? are you looking at the difference in file change in the file system and not just the file anymore?

test/s3cleanup/s3cleanup_test.go
jfantinhardesty: This is not really a test, in the sense of a test. This is to model what is done in the  other cleanup folder which deletes and clears out the buckets in Azure blob and data lake storage. This just mimics that in order to delete every object in our bucket. This is needed so that we can be sure that the size tracker is working correctly, as we assume an empty bucket at the start, and if there is a restart, that every object written was already written  by us.

test/e2e_tests/statfs_linux.go
jfantinhardesty: This is used in our tests to get the size returned by cloudfuse. The size that we return for the total size of all objects in our file systems is returned when the os makes a statfs call. So this is just a function to make that call and then convert that into the total number of bytes.

Dabnsky:
overall, pretty cool stuff. one thing I ran into when I ran it was this error: 

Error:  failed to initialize new pipeline [config error in Pipeline [component size_tracker is out of order]]

so a naïve user who just tacks that on their config file won't know that there is a certain order that it needs to be listed in the components section. we can consider adding sanitation support for the config file when it is being processed.

Great job!

jfantinhardesty:
> overall, pretty cool stuff. one thing I ran into when I ran it was this error:
> 
> Error: failed to initialize new pipeline [config error in Pipeline [component size_tracker is out of order]]
> 
> so a naïve user who just tacks that on their config file won't know that there is a certain order that it needs to be listed in the components section. we can consider adding sanitation support for the config file when it is being processed.
> 
> Great job!

Yes, good point. Yeah there is an implicit order in the component section, and this does require being after the file_cache section and before the attr_cache section. Once merged, I would update the current documentation on this that exists on our wiki at https://github.com/Seagate/cloudfuse/wiki/Config-File where we note the order of the configuration file and the priority that should be set if users want to use this component. Though, this component should really only be used by the plugin most likely.

Dabnsky (approved):

PR #425, opened 2025-01-24
Fix golang 1.23.4

jfantinhardesty:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Previously there was an issue upgrading past go version 1.23.2 where on mount we would get an error that the application failed to daemonize, but it actually started correctly. This fixes that issue.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


cmd/mount_linux.go
```diff
@@ -70,10 +70,10 @@ func createDaemon(pipeline *internal.Pipeline, ctx context.Context, pidFileName
 		log.Err("mount : failed to daemonize application [%v]", err)
 		return Destroy(fmt.Sprintf("failed to daemonize application [%s]", err.Error()))
 	}
+	defer dmnCtx.Release()
```
foodprocessor, line 73: From the library README, I'm gathering that `Release()` deletes the pid file. Only the child should do that, on exit, right?

foodprocessor, line 73: We could probably avoid the lint error, and improve readability besides, if we used this (from the daemon README):
```
defer func() {
	if err := context.Release(); err != nil {
		log.Printf("Unable to release pid-file: %s", err.Error())
	}
}()
```

foodprocessor (changes requested):
I'm out of my depth, but this is interesting!

cmd/mount_linux.go
jfantinhardesty: Yep, you are completely right. I was messing around with some changes trying to fix it and it looks like this accidently got included.

foodprocessor (approved):

PR #429, opened 2025-01-29
Discern cloud component

Dabnsky:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This defaults the mount to an s3 bucket if cloudfuse finds valid credentials in the environment when the config file does not have sufficient information regarding the cloud provider. if not s3, it will try the same thing for azstorage.
### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


cmd/mount.go
```diff
@@ -352,7 +352,19 @@ var mountCmd = &cobra.Command{
 				pipeline = append(pipeline, "attr_cache")
 			}
 
-			pipeline = append(pipeline, "s3storage")
+			// TODO: refer to the getBucketListS3() that mount_all does and getContainerListAzure() and let the restuls determine path.
+
```
foodprocessor, line 356: Outdated TODO

cmd/mount.go
```diff
@@ -352,7 +352,19 @@ var mountCmd = &cobra.Command{
...
+			containers, err := getBucketListS3()
+			if len(containers) != 0 && err == nil {
+				pipeline = append(pipeline, "s3storage")
+			} else if containers, err = getContainerListAzure(); err == nil {
+				if len(containers) != 0 && err == nil {
```
foodprocessor, line 361: There's a mix here, of an inline condition, and a separate, partially redundant condition.
I see why the inline condition makes sense here, and I agree with its use. But then the whole thing should be on one line.

cmd/mount.go
```diff
@@ -352,7 +352,19 @@ var mountCmd = &cobra.Command{
...
+					pipeline = append(pipeline, "azstorage")
+				}
+			} else if err != nil {
+				return fmt.Errorf("failed to mount instance [%s]", err.Error())
```
foodprocessor, line 365: Can this error message provide more context? Maybe mention that the pipeline components were not specified in config, and that automatically detecting the right cloud backend failed?

foodprocessor (changes requested):
Looks pretty good!

jfantinhardesty (changes requested):
Looks good. Also, this broke some of the unit tests because some of the error messages have changed. Could you fix those unit tests?

cmd/mount.go
```diff
@@ -354,7 +354,14 @@ var mountCmd = &cobra.Command{
...
 			}
 
-			pipeline = append(pipeline, "s3storage")
+			if containers, err := getBucketListS3(); len(containers) != 0 && err == nil {
+				pipeline = append(pipeline, "s3storage")
+			} else if containers, err = getContainerListAzure(); len(containers) != 0 && err == nil {
+				pipeline = append(pipeline, "azstorage")
+			} else if err != nil {
+				return errors.New("failed to initialize new pipeline :: Config does not specify any pipeline components and couldn't determine the cloud provider: " + err.Error())
+			}
```
foodprocessor, line 363: Maybe this should just be `else` rather than `else if ...`.
That way you report an error for a failure where the credentials work but there is no bucket to connect to.

foodprocessor (changes requested):
We should probably make another change so cloudfuse works without passing the `--config-file` switch. `TestForegroundConfigFileNotProvided` tests this (currently expecting it to fail).

foodprocessor (approved):
Looks good!

jfantinhardesty (approved):
Looks good

PR #431, opened 2025-02-01
Improve performance Size Tracker

jfantinhardesty:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This change improves the performance of the size tracker by moving the file IO out of the main thread and using a separate go routine to update the file every 10 seconds. This also moves the size to use an atomic data structure. This should improve performance of the size tracker by not locking the size on every interaction with the filesystem and should make changes later to include backing this file to the cloud easier to manage.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


component/size_tracker/journal.go
```diff
@@ -59,55 +64,75 @@ func CreateSizeJournal(filename string) (*MountSize, error) {
...
 }
 
-func (s *MountSize) Add(delta uint64) (uint64, error) {
-	s.mu.Lock()
-	defer s.mu.Unlock()
-	s.size += delta
-	err := s.writeSizeToFile()
-	return s.size, err
+func (ms *MountSize) Add(delta uint64) uint64 {
+	return ms.size.Add(delta)
```
foodprocessor, line 114: Could CompareAndSwap protect from going negative here?

foodprocessor (changes requested):
Great! Except negatives.

component/size_tracker/journal.go
jfantinhardesty: Using a CAS loop, this should prevent negatives while preventing race conditions.

foodprocessor: Great!

foodprocessor (approved):

PR #439, opened 2025-02-07
Add Empty Directory Marker Support in S3

jfantinhardesty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Add support for an empty directory marker in S3storage. Support is off by default, to follow semantic versioning for now, but it is enable in all of our tests.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


foodprocessor (commented):
I see that a big part of what is changing here is our internal convention for object naming. This change allows a trailing slash to be part of our internal name for an object. I feel unsure about that choice - since we already support directory markers in Azure without doing this.
No platforms support having a file and a folder with the same name in the same directory, so it seems like we would be fine to store an object's directory flag as a separate flag until we get to s3wrappers.
If we do move forward with the change to allow trailing slashes in our internal object names, then I think there's some confusion here, about where trailing slashes are required, expected, allowed, and not allowed. If we keep this approach, maybe it would be best to sort out exactly where which convention is followed, to get consistency and prevent errors and redundant calls.

jfantinhardesty:
> I see that a big part of what is changing here is our internal convention for object naming. This change allows a trailing slash to be part of our internal name for an object. I feel unsure about that choice - since we already support directory markers in Azure without doing this. No platforms support having a file and a folder with the same name in the same directory, so it seems like we would be fine to store an object's directory flag as a separate flag until we get to s3wrappers. If we do move forward with the change to allow trailing slashes in our internal object names, then I think there's some confusion here, about where trailing slashes are required, expected, allowed, and not allowed. If we keep this approach, maybe it would be best to sort out exactly where which convention is followed, to get consistency and prevent errors and redundant calls.

Good point. The reason I went with this method of changing the name is because of a difference between Azure blob storage and S3. In the block_blob code when they write a directory they set a flag in the blob metadata for "hdi_folder" to true, which indicates that this file is actually representing a folder. In datalake storage, the Azure API has a function NewDirectoryClient which will create a client with the name of the directory. It correctly handles creating that with a trailing slash and representing it as a directory in Azure. 

Now for S3, the only way we can represent a directory is an empty object that ends in a trailing slash. This means that  for the directory operations, we need to add a trailing slash to the name when those come into s3storage. We could do it as a metadata flag, but that is not how Lyve Cloud represents empty folders. They use a trailing slash approach and I think some other S3 vendors do trailing slashes as well. So that seems to be the approach that we should take. One complication with that that I uncovered is that our JoinUnixFilepath function actually removes trailing slashes when that function is called. This is fine in Azure since they don't want any function with a trailing slash, but it causes problems for us.

I will work to cleanup this convention. I propose the following. For every directory operation (CreateDir, DeleteDir, IsDIrEmpty, StreamDir, RenameDir) in s3storage we will add a trailing slash to the end of the name at the start of those functions. For operations that include both files and directories (GetAttr), we will handle looking up the file regularly, and then like we currently do attempt to look up a directory name with a trailing slash if that fails.

Ferelith-maker (approved):

foodprocessor:
> Now for S3, the only way we can represent a directory is an empty object that ends in a trailing slash.

I was talking about our internal object name (the one the components pass around), whereas I think you're talking about the convention inside the storage service itself.
My suggestion was to use a flag in our `objAttr` struct to signal that an object is a directory, and then translate that to a trailing slash inside the s3storage component. I am not suggesting you change anything about how you communicate the concept of an empty folder to the storage provider itself.

PR #449, opened 2025-02-13
Correct validateStorageError in file_cache

foodprocessor:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

`validateStorageError()` is meant to identify whether an `ENOENT` error returned by a cloud call is due to the file being in cache but not in cloud (case 2), and to clear the error in that case.
`validateStorageError()` was incorrectly clearing cloud `ENOENT` errors whenever `createEmptyFile` was true (without checking whether the local file exists).
To correct this, we could either not check the `createEmptyFile` flag at all, or we should require it be false before clearing cloud `ENOENT` errors. I chose the second option.

I also set `recoverable` to true in a `RenameFile()` call to `validateStorageError()` where it was mistakenly left set to false (after changing `RenameFile()` to support renaming open files).

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -860,7 +860,7 @@ func (fc *FileCache) CreateFile(options internal.CreateFileOptions) (*handlemap.
 func (fc *FileCache) validateStorageError(path string, err error, method string, recoverable bool) error {
 	// For methods that take in file name, the goal is to update the path in cloud storage and the local cache.
 	// See comments in GetAttr for the different situations we can run into. This specifically handles case 2.
-	if !fc.createEmptyFile && os.IsNotExist(err) {
+	if os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


jfantinhardesty (approved):
Is there a unit test you could add to check for this?

foodprocessor:
> Is there a unit test you could add to check for this?

I feel like this bug is so obscure (since, for no discernible reason, it was only exposed when `createEmptyFile` is true), that creating a test for it would seem overly specific. Almost like checking that renaming a file still works as expected when it's 3pm.

Dabnsky (approved):

foodprocessor:
After reviewing this again, I think this change is unnecessary. I'm going to close this, at least for now.

PR #450, opened 2025-02-13
Allow GUI to only mount as a drive letter on Windows

jfantinhardesty:

- [ ] Refactor
- [x] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This changes the GUI to only allow mounting as a drive letter on Windows, to reduce user confusion and simplify the Windows mounting process. It replaces the browse button with one for drive letters that display the available drive letters to mount on. The old code is still there, so you can manually type a path and still mount to it, but the browse button in gone.

Here is how the GUI looks now on Windows.

image

image

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/gui/mountPrimaryWindow.py
+++ b/gui/mountPrimaryWindow.py
@@ -71,20 +71,15 @@ class FUSEWindow(settingsManager,configFuncs, QMainWindow, Ui_primaryFUSEwindow)
             #   https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#file-and-directory-names
             # Disallow the following [<,>,.,",|,?,*] - note, we still need directory characters to declare a path
             self.lineEdit_mountPoint.setValidator(QtGui.QRegularExpressionValidator(r'^[^<>."|?\0*]*$',self))
-            self.button_driveLetter = QtWidgets.QPushButton('Drive Letter')
-            self.button_driveLetter.setToolTip('Select an unused drive letter for mounting')
-
-            self.horizontalLayout_3.addWidget(self.button_driveLetter)
-            self.button_driveLetter.clicked.connect(self.chooseDriveLetter)
+            self.button_browse.setText('Drive Letter')
+            self.button_browse.setToolTip('Select an unused drive letter for mounting')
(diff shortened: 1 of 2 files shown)


foodprocessor (approved):

foodprocessor:
Can you push a screenshot of this in action into this PR's description so we can see it without having to build it?

gui/mountPrimaryWindow.py
```diff
@@ -69,13 +71,23 @@ def __init__(self):
             #   https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#file-and-directory-names
             # Disallow the following [<,>,.,",|,?,*] - note, we still need directory characters to declare a path
             self.lineEdit_mountPoint.setValidator(QtGui.QRegularExpressionValidator(r'^[^<>."|?\0*]*$',self))
+            self.button_driveLetter = QtWidgets.QPushButton('Drive Letter')
+            self.button_driveLetter.setToolTip('Select an unused drive letter for mounting')
+
+            self.horizontalLayout_3.addWidget(self.button_driveLetter)
```
Ferelith-maker, line 77: I would recommend renaming the horizontalLayout to something that can help identify where the dev is in the code with reference to the UI. For example, lineEdit_mountPoint tells us specifically what we're using and for what. Is there a way we can reference horizontalLayout_3 to something we are familiar with?

Ferelith-maker (changes requested):

gui/mountPrimaryWindow.py
Ferelith-maker: I realize now that a widget is being added after the UI has been compiled into a python file. Most of the elements have been added using the .ui source in QTCreator. What are you opinions in keeping to that convention, that way we don't have to rename horizontalLayout_3?

jfantinhardesty: Great suggestions, these have now been fixed. The button is defined in the GUI and the text and tooltip are set in the python code.

Ferelith-maker (approved):
Awesome!

PR #457, opened 2025-02-20
Support empty dir s3

jfantinhardesty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Add support for an empty directory marker in S3storage. Support is off by default, to follow semantic versioning for now, but it is enable in all of our tests.

This uses feedback from Michael on the previous PR and keeps all changes in the s3storage component by changing the getKey function to append a trailing slash if the object is a directory.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


component/s3storage/s3storage.go
```diff
@@ -223,7 +223,8 @@ func formatListDirName(path string) string {
 
 func (s3 *S3Storage) IsDirEmpty(options internal.IsDirEmptyOptions) bool {
 	log.Trace("S3Storage::IsDirEmpty : %s", options.Name)
-	list, _, err := s3.storage.List(formatListDirName(options.Name), nil, 1)
+	// List up to two objects, since one could be the directory with a trailing slash
+	list, _, err := s3.storage.List(formatListDirName(options.Name), nil, 2)
```
foodprocessor, line 227: The code below still only checks if the number of objects found is zero. I think we need more logic there.

component/s3storage/s3wrappers.go
```diff
@@ -245,6 +252,23 @@ func (cl *Client) copyObject(source string, target string, isSymLink bool) error
...
+	// in this case we don't need to check if the file exists, so we use deleteObject, not DeleteFile
+	// this is what S3's DeleteObject spec is meant for: to make sure the object doesn't exist anymore
+	err = cl.deleteObject(source, isSymLink, isDir)
+	if err != nil {
+		log.Err("Client::renameObject : deleteObject(%s) failed. Here's why: %v", source, err)
+	}
+
+	return err
+}
+
```
foodprocessor, line 271: My thinking about this file, is that it only has functions which wrap AWS SDK calls.
I'm not necessarily against having this one here, but it does buck that trend.

component/s3storage/client.go
```diff
@@ -307,13 +307,18 @@ func (cl *Client) CreateFile(name string, mode os.FileMode) error {
...
-	// directories will be represented only as object prefixes
-	// we have no way of representing an empty directory, so do nothing.
-	// Note: we could try to list the directory and return EEXIST if it has contents,
-	// but that would be a performance penalty for a check that the OS already does.
+
+	// If the S3 endpoint does not support directory markers then we can do nothing here.
 	// So, let's make it clear: we expect the OS to call GetAttr() on the directory
 	// to make sure it doesn't exist before trying to create it.
+	if cl.Config.enableDirMarker {
+		err := cl.putObject(name, nil, 0, false, true)
```
foodprocessor, line 315: I'm fine with it the way it is, but I think the number of arguments here is pushing the readability limit. If you'd like to use an options struct here (and in other similar spots), I'd support that.

foodprocessor (changes requested):
Looks great!
I think just IsDirEmpty needs a bit of love.

component/s3storage/client.go
jfantinhardesty: Good idea. I added structs for some of the functions that had a lot of arguments in s3wrappers.

component/s3storage/s3wrappers.go
jfantinhardesty: Yes it does a bit, but it is kind of a wrapper for some of the other s3 wrappers, so it seems like it should belong near it.

component/s3storage/s3storage.go
jfantinhardesty: The current logic should work. Calling list with a count of 2 allows for a trailing slash to be included since it will list the prefix, and then possibly the directory name with a trailing slash. Since list removes the current directory from the list, the list returned should still be empty, regardless of whether there are directory markers or not. I added an additional test for IsDirEmpty to cover the case when they are not using directory markers.

foodprocessor (approved):

Ferelith-maker (approved):
Nice job

PR #490, opened 2025-03-25
Add commands on Windows to remount as either user or system

jfantinhardesty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Adds a new command option on Windows to remount the mount as either the user or the system. Using the system will require elevated permissions.

### Checklist

- [ ] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/cmd/service_windows.go
+++ b/cmd/service_windows.go
@@ -136,6 +136,11 @@ var uninstallCmd = &cobra.Command{
 			return fmt.Errorf("Failed to remove as a Windows service. Here's why: %v", err)
 		}
 
+		err = deleteMountTracker()
+		if err != nil {
+			return fmt.Errorf("Failed to remove mount.json tracker file. Here's why: %v", err)
+		}
+
 		return nil
 	},
 }
@@ -263,6 +268,10 @@ func removeService() error {
 		return err
 	}
 	return nil
+}
+
+func deleteMountTracker() error {
+	return winservice.DeleteMountJSONFile()
 }
 
 func init() {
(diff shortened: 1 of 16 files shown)


cmd/service_windows.go
```diff
@@ -143,40 +180,100 @@ var removeRegistryCmd = &cobra.Command{
...
+	err = service.Delete()
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
+func deleteMountTracker() error {
+	return winservice.DeleteMountJSONFile()
+}
```
foodprocessor, line 275: This function is so small... can we drop it and just have the caller call this function directly?

internal/winservice/mount_tracker.go
```diff
@@ -125,10 +135,31 @@ func removeMount(mounts Mounts, mountPath string) Mounts {
 	return filteredMounts
 }
 
+// DeleteMountJSONFile deletes all the mount.json files on the system
+func DeleteMountJSONFile() error {
```
foodprocessor, line 139: Should this be plural now?

tools/windows-service/main.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+		}
+	}
+}
+
+func main() {
+	isService, err := svc.IsWindowsService()
+	if err != nil || !isService {
+		log.Err("Unable to determine if running as Windows service or not running as Windows service: %v", err.Error())
+	}
```
foodprocessor, line 132: Is there a reason you decided not to enforce this check?
I'm guessing that the risk this would protect against is that a user could launch the executable directly, and then it could misbehave (run multiple times, or something might be wrong with its environment)?

foodprocessor (approved):
Looks good!

tools/windows-service/main.go
jfantinhardesty: Nope, just forgot to. Good catch, and I think we probably should enforce it.

PR #491, opened 2025-03-27
Password is passphrase

jfantinhardesty:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This changes the passphrase field to be an actual password. It uses argon2id to convert that to a key which is then used for the aes encryption of the config file. This should make it easier to use since you can actually just provide a standard password, rather than a base64 encoded aes encryption key.

### Checklist

- [x] Tested locally
- [x] Added new dependencies
- [x] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/cmd/secure_test.go
+++ b/cmd/secure_test.go
@@ -209,7 +209,7 @@ func (suite *secureConfigTestSuite) TestSecureConfigEncryptInvalidKey() {
 	defer suite.cleanupTest()
 	confFile, _ := os.CreateTemp("", "conf*.yaml")
 	outFile, _ := os.CreateTemp("", "conf*.yaml")
-	passphrase := base64.StdEncoding.EncodeToString([]byte("123"))
+	passphrase := base64.StdEncoding.EncodeToString([]byte(""))
 
 	defer os.Remove(confFile.Name())
 	defer os.Remove(outFile.Name())
--- a/cmd/secure_test.go
+++ b/cmd/secure_test.go
@@ -27,7 +27,6 @@ package cmd
 
 import (
 	"bytes"
-	"encoding/base64"
 	"fmt"
 	"os"
 	"testing"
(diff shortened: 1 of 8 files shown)


cmd/secure_test.go
```diff
@@ -209,7 +209,7 @@ func (suite *secureConfigTestSuite) TestSecureConfigEncryptInvalidKey() {
 	defer suite.cleanupTest()
 	confFile, _ := os.CreateTemp("", "conf*.yaml")
 	outFile, _ := os.CreateTemp("", "conf*.yaml")
-	passphrase := base64.StdEncoding.EncodeToString([]byte("123"))
+	passphrase := base64.StdEncoding.EncodeToString([]byte(""))
```
foodprocessor, line 212: You can drop the base64 encoding here, right?

common/encryption.go
```diff
@@ -0,0 +1,39 @@
...
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+   SOFTWARE
+*/
+
+package common
+
+import "golang.org/x/crypto/argon2"
+
+const (
+	SaltLength    = uint16(16)
```
foodprocessor, line 30: 16 bytes of salt seems like a lot.

foodprocessor (approved):
Wow, this is fun!

common/encryption.go
foodprocessor: I'm reading that 16 the currently accepted recommendation. That's crazy to me. I feel like if a cryptographic hash is worth its salt (get it?), then it shouldn't need much salt.

jfantinhardesty: 16 bytes is actually the recommended amount for password hashing. Here is the RFC on argon2id. https://datatracker.ietf.org/doc/rfc9106/. The Go implementation and discussion of defaults is also here https://pkg.go.dev/github.com/alexedwards/argon2id#section-readme.

Ferelith-maker (approved):

PR #514, opened 2025-04-09
Update windows

jfantinhardesty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Add support for the update command on Windows. It will automatically pop up the menu to get admin rights to install the machine. It also installs while cloudfuse is running and Inno Setup will then replace the executable to the new one on a system restart.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


cmd/update.go
```diff
@@ -148,8 +169,39 @@ func hasCommand(command string) bool {
...
+		"/VERYSILENT",
+		"/NOICONS",
+		"/NORESTART",
+		"/SUPPRESSMSGBOXES",
+		"/dir=expand:{autopf}\\Cloudfuse",
+	}
+
+	cmd := exec.Command(absPath, args...)
+
+	// cmd.Stdout = os.Stdout
```
Dabnsky, line 190: do we need these comments?

cmd/update.go
```diff
@@ -75,27 +77,36 @@ var updateCmd = &cobra.Command{
...
+			if opt.Package != "tar" && opt.Package != "deb" && opt.Package != "rpm" {
+				return errors.New("--package should be one of tar|deb|rpm")
+			}
+			if os.Geteuid() != 0 && opt.Output == "" && (opt.Package == "deb" || opt.Package == "rpm") {
+				return errors.New(".deb and .rpm requires elevated privileges")
+			}
+			if opt.Output == "" && opt.Package == "tar" {
+				return errors.New("Need to pass parameter --package with deb or rpm or pass parameter --output with location to download to")
+			}
+
```
Dabnsky, line 99: I don't see installUpdate() being called in the linux case. only for windows

Dabnsky (changes requested):
Looks great! Just need to call the same installUpdate function for linux

cmd/update_test.go
```diff
@@ -144,6 +144,75 @@ func (suite *updateTestSuite) TestUpdateWithOutputTarLinux() {
...
+
+	_, err = executeCommandC(rootCmd, "update", "--package=deb", fmt.Sprintf("--output=%s", outputFile.Name()))
+	suite.assert.Error(err)
+
+	_, err = executeCommandC(rootCmd, "update", "--package=rpm", fmt.Sprintf("--output=%s", outputFile.Name()))
+	suite.assert.Error(err)
+
+	os.Remove(outputFile.Name())
+}
+
```
Dabnsky, line 215: I think we need to add a test to make sure the linux and windows updates work with correct inputs

cmd/update.go
jfantinhardesty: Nope. Good catch

cmd/update.go
jfantinhardesty: It is called in both. After the switch case, both Linux and Windows will call installUpdate.

cmd/update.go
foodprocessor, line 1: I know this isn't Windows-specific, but can you change the description of the "package" option to say the default is to automatically detect it?

cmd/update.go
```diff
@@ -75,31 +77,40 @@ var updateCmd = &cobra.Command{
...
+		switch runtime.GOOS {
+		case "linux":
+			if opt.Package != "tar" && opt.Package != "deb" && opt.Package != "rpm" {
+				return errors.New("--package should be one of tar|deb|rpm")
+			}
+			if os.Geteuid() != 0 && opt.Output == "" && (opt.Package == "deb" || opt.Package == "rpm") {
+				return errors.New(".deb and .rpm requires elevated privileges")
+			}
+			if opt.Output == "" && opt.Package == "tar" {
+				return errors.New("need to pass parameter --package with deb or rpm or pass parameter --output with location to download to")
```
foodprocessor, line 97: This sentence has two "or"s pretty close together. Maybe a comma would help.
The same goes for the same error message in the Windows block below.

cmd/update.go
foodprocessor, line 1: Also not Windows specific...
Should we use `NewRequestWithContext` instead of `http.Get` in the `downloadUpdate` function?
Or is `http.Get` just better at handling the large binary payload?

foodprocessor, line 1: For Linux, this uses dpkg -i command to update... Is there any reason to use apt for this? We had to for a fresh install to install the fuse dependency, so I guess what I'm asking is: is there any risk that we end up adding a new dependency as part of a future update?

foodprocessor (approved):
I have some comments, but none of them are about the new Windows part you added, which looks great!

cmd/update.go
jfantinhardesty: Yeah, that is much clearer

cmd/update.go
jfantinhardesty: Fixed

cmd/update.go
jfantinhardesty: Yep, just oversite on my part.

jfantinhardesty: Yeah, it is probably better to use apt here. I will change that.

cmd/update.go
Dabnsky: yep, I think I got confused from the diff view I was using. Sorry. Looks good

Dabnsky (approved):

PR #515, opened 2025-04-09
Log collector cmd

Dabnsky:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This adds the dumpLogs command for cloudfuse to collect logs into an archive

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


cmd/log-collector.go
```diff
@@ -0,0 +1,300 @@
...
+
+var dumpPath string
+var logConfigFile string
+var dumpLogsCmd = &cobra.Command{
+	Use:               "dumpLogs",
+	Short:             "interface to gather and review cloudfuse logs",
+	Long:              "interface to gather and review cloudfuse logs",
+	SuggestFor:        []string{"dump", "dumpLog", "dumpLogs"},
+	Args:              cobra.ExactArgs(1),
+	Example:           "cloudfuse dumpLogs ",
```
jfantinhardesty, line 56: I would suggest changing this from accepting an argument, to just dumping the zip of the logs in the current directory. Then you do not need ExactArgs to be 1. But maybe others have thoughts on how this should function.

cmd/log-collector.go
```diff
@@ -0,0 +1,300 @@
...
+			err = createArchive(filteredSyslogPath) //supply the path of the filtered log file here
+			if err != nil {
+				return fmt.Errorf("unable to create archive: [%s]", err.Error())
+			}
+		}
+
+		// are any 'base' logging or syslog filters being used to redirect to a separate file?
+		// check for /etc/rsyslog.d and /etc/logrotate.d files
+
+		return fmt.Errorf("failed to collect logs and create archive [%s]", err.Error())
```
jfantinhardesty, line 141: I think some of the logic in this case is incorrect. It should be the case that if you make it here, it returns successfully, early returns should be the failure path, typically. In this case, I got a runtime error when I ran this code because I was using base on Linux, which then does not product an error and ends up here.

cmd/log-collector.go
```diff
@@ -0,0 +1,300 @@
...
+			}
+		}
+	}
+	writer.Flush()
+
+	return outPath, scanner.Err()
+}
+
+func createArchive(logPath string) error {
+	ArchiveName := fmt.Sprintf("cloudfuse_Logs_" + time.Now().Format("2006-01-02_15-04-05"))
```
jfantinhardesty, line 186: Fix the lint error here.

cmd/log-collector.go
```diff
@@ -0,0 +1,300 @@
...
+				return "", err
+			}
+		}
+	}
+	writer.Flush()
+
+	return outPath, scanner.Err()
+}
+
+func createArchive(logPath string) error {
```
jfantinhardesty, line 185: This is a somewhat large function, could you split this into two, one for Linux and one for Windows.

cmd/log-collector.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+	}
+
+	return err
+}
+
+func init() {
+	rootCmd.AddCommand(dumpLogsCmd)
+	// dumpLogsCmd.Flags().StringVar(&dumpPath, "dump-path", "", "Input archive creation path")
```
jfantinhardesty, line 297: You could cleanup these commands.

cmd/log-collector.go
```diff
@@ -0,0 +1,300 @@
...
+	"github.com/spf13/cobra"
+)
+
+var dumpPath string
+var logConfigFile string
+var dumpLogsCmd = &cobra.Command{
+	Use:               "dumpLogs",
+	Short:             "interface to gather and review cloudfuse logs",
+	Long:              "interface to gather and review cloudfuse logs",
+	SuggestFor:        []string{"dump", "dumpLog", "dumpLogs"},
```
jfantinhardesty, line 54: I'm not sure I love the name, mainly because it requires the L to be capitalized. Not sure if anyone else had ideas for a name.

cmd/log-collector.go
```diff
@@ -0,0 +1,300 @@
+//go:build linux
```
jfantinhardesty, line 1: You have code for Windows here, but the build flag will only build for Linux.

cmd/log-collector.go
```diff
@@ -0,0 +1,300 @@
...
+			filteredSyslogPath, err := createFilteredLog(logPath) //generate a separate .log file and place it in a folder. output the path of the filtered log file
+			if err != nil {
+				return fmt.Errorf("failed to crate a filtered log from the syslog: [%s]", err.Error())
+			}
+			filteredSyslogPath = filepath.Dir(filteredSyslogPath)
+			err = createArchive(filteredSyslogPath) //supply the path of the filtered log file here
+			if err != nil {
+				return fmt.Errorf("unable to create archive: [%s]", err.Error())
+			}
+		}
```
jfantinhardesty, line 136: If you do syslog and Windows it should give an error that you need to view the logs directly in the event viewer. And in the case the user sets the logs to silent, it should give an error as well.

cmd/log-collector.go
```diff
@@ -0,0 +1,300 @@
...
+
+		tarWriter := tar.NewWriter(gzWriter)
+		defer tarWriter.Close()
+
+		items, err := os.ReadDir(logPath)
+		if err != nil {
+			return err
+		}
+
+		for _, item := range items {
```
jfantinhardesty, line 208: You dump everything from the log folder to the logs, but they might not be logs, and in many cases there will be other files there. You should follow the logic for how the log files get written. In the logging config, there is a filepath that gets set, then there is a file-count, where a 1, 2, 3, etc. gets appended to the name once the old file files up. It should add these files and only these files.

jfantinhardesty (changes requested):
I like the feature and it is a good start, just some comments to make it work better for base logging and some minor issues. Also, do you think you could write some unit tests for this as well?

cmd/log-collector.go
Dabnsky: I was anticipating further discussion for a proper name

cmd/log-collector.go
Dabnsky: I see IBM and mysql both have a default location to store logs and they also support an option to specify an output location other than the default by using a flag. I'll move forward with this approach.

Thank you

cmd/log-collector.go
Dabnsky: These are now uncommented after reviewing other log collection cmd tools.

cmd/log-collector.go
foodprocessor, line 1: I think this could really use some code comments to answer what each section of code trying to accomplish, and what assumptions are being made.

cmd/log-collector.go
Dabnsky: here are my sources: 

IBM: 
    https://www.ibm.com/docs/en/taddm/7.3.0?topic=cis-running-collect-logs-command-collect-package-log-files
mysql:
    https://dev.mysql.com/doc/mysql-cluster-manager/1.4/en/mcm-collect-logs.html
    
   the IBM doc doesn't go over on specifying a directory as an option, but the mysql one does. I moved forward with the mysql exmaple.

PR #520, opened 2025-04-11
Fix case sensitive unmount

jfantinhardesty:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This address a case sensitive issue where previously mounts had to the the same case unmounted as when they were mounted. This fixes that so it is case insensitive.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


internal/winservice/service_windows.go
```diff
@@ -80,7 +81,7 @@ func StartMount(mountPath string, configFile string, passphrase *memguard.Enclav
 
 // StopMount stops the mount if the name exists in the WinFsp Windows registry.
 func StopMount(mountPath string) error {
-	instanceName := mountPath
+	instanceName := strings.ToLower(mountPath)
 
 	buf := writeCommandToUtf16(stopCmd, SvcName, instanceName, mountPath)
```
foodprocessor, line 86: Should we be at all concerned that the instance name could have been written by an older version of our software, and now it won't match? Should we list the mounts and lowercase them to find a match which might be missed otherwise?

foodprocessor (changes requested):
I think we also need to change `removeMount` in mount_tracker.go.

internal/winservice/service_windows.go
```diff
@@ -51,36 +52,35 @@ const (
...
-	}
-
-	instanceName := mountPath
+	instanceName := strings.ToLower(mountPath)
 
-	if passphrase != nil {
-		buff, err := passphrase.Open()
-		if err != nil || buff == nil {
-			return errors.New("unable to decrypt passphrase key")
+	err := startMountHelper(instanceName, mountPath, configFile, passphrase)
```
jfantinhardesty, line 57: Add comments here

foodprocessor (approved):

Ferelith-maker (approved):

PR #527, opened 2025-04-21
Update blobfuse2.4.1

jfantinhardesty:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This updates to the version 2.4.1 of blobfuse2. Here are some of the highlights of what they have added in this version.

Features
- Mount container or directory but restrict the view of blobs that you can see. This feature is available only in read-only mount.
- To protect against accidental overwrites on data stored by block-cache on temp path, crc64 hash will be validated on read. This feature can be enabled by using --block-cache-strong-consistency cli flag.
- To provide strong consistency check, ETAG of the file will be preserved on open. For any subsequent block download, with block-cache, ETAG will be verified and if the blob has changed in container the download will be declare failure resulting into read failure.

Bug Fixes
- Do not wipeout temp-cache on start after a un-graceful unmount, if cleanup-on-start is not configured in file-cache.
- When the subdirectory is mounted and there is some file/folder operation, remove only the subdirectory path from the file paths.
- Enable atomic_o_trunc flag in libfuse to allow O_TRUNC flag to come in the open call for fuse2.
- In file-cache, when the O_TRUNC flag is passed to the open call and no modifications were done to the file before closing it then update the file in the Azure Storage to size 0.

Other Changes
- Optimized listing operation on HNS account to support symlinks.
- Optimized Rename operation to do less number of REST calls.
### Checklist

- [x] Tested locally
- [x] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


foodprocessor (approved):
Looks good!
Thank you!

PR #556, opened 2025-05-13
Network Fault Tolerance

foodprocessor:
The purpose of this PR is to start getting CI test results, and to open this WIP to feedback.
This is a draft PR for a feature to allow offline access.
At this point, requests that match our offline access policy are allowed to act on files in the file cache, and we return success. But we do not service those operations once the connection is reestablished, although we do block eviction for files which are awaiting cloud synchronization.
File synchronization, journaling, unit testing, and end-to-end testing are gating features before this can come out of draft.


component/attr_cache/attr_cache.go
```diff
@@ -447,15 +451,25 @@ func (ac *AttrCache) StreamDir(
...
-	}
-	// add cached items in
-	if len(cachedPathList) > 0 {
-		log.Info(
-			"AttrCache::StreamDir : %s merging in %d list cache entries...",
-			options.Name,
-			len(cachedPathList),
-		)
-		pathList = append(pathList, cachedPathList...)
+	} else if errors.Is(err, &common.CloudUnreachableError{}) {
```
jfantinhardesty, line 454: Do you still need to handle the regular case of appending the cached list to the path list?

component/attr_cache/attr_cache.go
```diff
@@ -586,16 +607,19 @@ func (ac *AttrCache) cacheListSegment(
 	}
 	// add the new entry
 	listDirItem.listCache[token] = newListCacheSegment
-	// scan the listing cache and remove expired entries
```
jfantinhardesty, line 589: Is this deleted so that the expiration of entries is solely handled by the LRU, so you don't have to check the offline files?

component/file_cache/file_cache.go
```diff
@@ -596,7 +740,7 @@ func (fc *FileCache) StreamDir(
 				// and container or not. So we rely on getAttr to tell if entry was cached then it exists in cloud storage too
 				// If entry does not exists on storage then only return a local item here.
 				_, err := fc.NextComponent().GetAttr(internal.GetAttrOptions{Name: entryPath})
-				if err != nil && (err == syscall.ENOENT || os.IsNotExist(err)) {
+				if err != nil && errors.Is(err, os.ErrNotExist) {
```
jfantinhardesty, line 743: Does ErrNotExist check for ENOENT. I thought that the error.Is did not match against syscall errors?

component/file_cache/file_cache.go
```diff
@@ -757,8 +899,15 @@ func (fc *FileCache) RenameDir(options internal.RenameDirOptions) error {
 
 	// rename the directory in the cloud
 	err = fc.NextComponent().RenameDir(options)
-	if err != nil {
-		log.Err("FileCache::RenameDir : error %s [%s]", options.Src, err.Error())
+	// if we are offline, and offline access is enabled, allow local directories to be renamed
+	if isOffline(err) && fc.offlineOperationAllowed(options.Src) && fc.notInCloud(options.Dst) {
+		log.Debug(
```
jfantinhardesty, line 904: Should this be a log.Warn?

jfantinhardesty (approved):

component/attr_cache/attr_cache.go
jfantinhardesty: Here I'm referring to the the original code which appended the cachedPathList to the pathList in all cases. In the change it appears that only happens when the Cloud is unreachable error happens. Shouldn't it still need to add cached items in the other case?

component/attr_cache/attr_cache.go
foodprocessor: My thought process was that if you lost your cloud connection in the middle of listing a directory whose contents did not change, I didn't want to lose the rest of the cached pages for the directory listing. That way, if the contents didn't change (and the tokens still match), the listing could actually complete offline.
In other words, this is an attempt to keep expired directory listing pages available instead of deleting them when they expire.
This is a bit of a memory leak. I could fix that by walking the pages and deleting the useless ones after the new listing completes. I'll make that a backlog item for now.

component/file_cache/file_cache.go
foodprocessor: In Go Playground `errors.Is(syscall.ENOENT, os.ErrNotExist)` returns true, so I think we're good.
I do want to note though, that `os.IsNotExist` does not work on joined errors, which is why I'm using the newer `errors.Is` interface.

component/file_cache/file_cache.go
foodprocessor: Yeah... it should be. Thanks!

component/attr_cache/attr_cache.go
foodprocessor: Ok, I see. You're talking about merging in `cachedPathList`.
That code block needed to be dropped regardless of this feature branch.

Much more detail:
I just did some digging, and found where that removed code actually comes from.
It was originally introduced in [this commit](https://github.com/Seagate/cloudfuse/pull/107/commits/2aa07d3d760a8ff5ea508abe2ccfa2da793de5d1), when we first started caching pages in StreamDir.
That original version was confused. It tried to accrue all the directory's contents into one giant ready-to-return list, instead of the map of pages of directory contents (`listCache`) which we have now.
So this idea, of merging a cached page with a new page from the cloud, was based on the now-defunct idea of returning more than 1000 entries in response to a single request.
In the context of the new code, merging a cached page with a fresh page is somewhere between useless and hazardous - it may cause us to double-list some objects.

PR #561, opened 2025-05-20
Log collector cmd

Dabnsky:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Adds a gatherLogs command to the cloudfuse CLI that collects log files into an archive

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #


cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+			logPath = filepath.Dir(logPath)
+			if runtime.GOOS == "linux" {
+				err = createLinuxArchive(logPath)
+				if err != nil {
+					return fmt.Errorf("unable to create archive: [%s]", err.Error())
+				}
+			} else if runtime.GOOS == "windows" {
+				//TODO: add the system app data system32 if there is no filepath in the config.
+				if strings.HasPrefix(logPath, ".cloudfuse") {
+					err = createWindowsArchive("path/to/cloudfuse/app/data/system32/thing")
```
jfantinhardesty, line 82: This should be the actual path.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/Seagate/cloudfuse/common"
+	"github.com/Seagate/cloudfuse/common/config"
+	"github.com/spf13/cobra"
+)
+
+var dumpPath string
```
jfantinhardesty, line 45: Could this be called outputPath instead? Since the flag is output-path.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+
+		return nil
+	},
+}
+
+// checkOutputPath makes sure the path for archive creation is valid.
+func checkOutputPath(outPath string) error {
+	var err error
+	if outPath == "" {
+		dumpPath, err = os.Getwd()
```
jfantinhardesty, line 118: Here dumpPath refers to the global variable dumpPath, which I don't think is the intention.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+			} else if logType == "base" {
+				if config.IsSet("logging.file-path") {
+					err = config.UnmarshalKey("logging.file-path", &logPath)
+					if err != nil {
+						return "", "", err
+					}
+					if strings.HasPrefix(logPath, "$HOME") {
+						logPath = common.ExpandPath(logPath)
+					}
+					logPath, err = filepath.Abs(logPath)
```
jfantinhardesty, line 173: Should return if err is not nil here.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+
+	//first check logPath is valid
+	_, err := os.Stat(logPath)
+	if err != nil {
+		return err
+	}
+
+	//setup tar.gz file
+	outFile, err := os.Create(dumpPath + "/cloudfuse_logs.tar.gz")
+	if err != nil {
```
jfantinhardesty, line 238: I would move the hardcode paths to the top of the file as a constant.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+}
+
+// createFilteredLog creates a new log file containing only the cloudfuse logs from the input log file.
+// It only runs for linux when the configured logging type is set to "syslog"
+func createFilteredLog(logFile string) (string, error) {
+
+	keyword := "cloudfuse"
+
+	os.Mkdir("/tmp/cloudfuseSyslog", 0760)
+
```
jfantinhardesty, line 196: Since there is a temporary file being made here, you could use os.MkdirTemp or os.CreateTemp to create temporary files and directories, rather than using one hard coded one.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+		logType, logPath, err := getLogInfo(logConfigFile)
+		if err != nil {
+			fmt.Errorf("failed to parse config file [%s]", err.Error())
+		}
+
+		if logType == "silent" {
+			return fmt.Errorf("no logs were generated due to log type being silent")
+		} else if logType == "base" {
+			logPath = filepath.Dir(logPath)
```
jfantinhardesty, line 73: I don't think this should set logPath to this for clarity. I would make the variable logPathDir.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+	//populate tar.gz file
+	var amountLogs int
+	items, err := os.ReadDir(logPath)
+	if err != nil {
+		return err
+	}
+	for _, item := range items {
+		if strings.HasPrefix(item.Name(), "cloudfuse") && strings.HasSuffix(item.Name(), ".log") {
+			itemPath := filepath.Join(logPath, item.Name())
+			itemPath = filepath.Clean(itemPath)
```
jfantinhardesty, line 258: Filepath join actually calls clean, so the clean after is not needed.

cmd/log-collector_test.go
```diff
@@ -0,0 +1,358 @@
...
+	suite.assert = assert.New(suite.T())
+	err := log.SetDefaultLogger("base", common.LogConfig{Level: common.ELogLevel.LOG_DEBUG()})
+	if err != nil {
+		panic(fmt.Sprintf("Unable to set debug logger as default: %v", err))
+	}
+}
+
+func (suite *logCollectTestSuite) cleanupTest(dir string) {
+	resetCLIFlags(*gatherLogsCmd)
+	os.Remove(dir + "/cloudfuse_logs.tar.gz")
```
jfantinhardesty, line 43: This should also cleanup the windows zip as well.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+			}
+		}
+	}
+	writer.Flush()
+
+	return outPath, scanner.Err()
+}
+
+func createLinuxArchive(logPath string) error {
+
```
jfantinhardesty, line 229: There should not be a leading empty line after a function. Is the linter running in your VS Code?

jfantinhardesty (changes requested):

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+	"github.com/Seagate/cloudfuse/common"
+	"github.com/Seagate/cloudfuse/common/config"
+	"github.com/spf13/cobra"
+)
+
+var dumpPath string
+var logConfigFile string
+var gatherLogsCmd = &cobra.Command{
+	Use:               "gatherLogs",
```
jfantinhardesty, line 48: Could the command be gather-logs instead. This way there is no capital and it conforms to other commands like gen-config.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+			return fmt.Errorf("could not use the output path %s, [%s]", dumpPath, err)
+		}
+
+		if logConfigFile, err = filepath.Abs(logConfigFile); err != nil {
+			return fmt.Errorf("couldn't determine absolute path for config file [%s]", err.Error())
+		}
+
+		logType, logPath, err := getLogInfo(logConfigFile)
+		if err != nil {
+			fmt.Errorf("failed to parse config file [%s]", err.Error())
```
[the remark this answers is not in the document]
foodprocessor: Doesn't the parameter parsing system assign the default path to this one for you? In that case, returning an error here makes sense.
If you do intend to return an error here, you are missing the `return` keyword.

cmd/log-collector.go
```diff
@@ -0,0 +1,362 @@
...
+		dumpPath, err = os.Getwd()
+		if err != nil {
+			return err
+		}
+	} else {
+		if !common.DirectoryExists(outPath) {
+			return err
+		}
+
+		dumpInfo, err := os.Stat(outPath)
```
foodprocessor, line 127: If you don't want to shadow `err` here, you have to drop the colon, which means you'll have to declare `dumpInfo` on a separate line. This is a weakness in Go due to its combining assignment-declarations and multiple-variable returns.

foodprocessor (changes requested):

PR #598, opened 2025-07-03
Default to first *authorized* bucket

foodprocessor:
Having `bucket-name` be optional was broken when `ListBuckets` would return buckets for which the credentials do not provide access. This adds a new function, `filterAccessibleBuckets()`, to call `HeadBucket` on the listed buckets in parallel, and identify the accessible one(s). Then we can choose a default which actually works.

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/cmd/mount_all.go
+++ b/cmd/mount_all.go
@@ -263,7 +263,7 @@ func getBucketListS3() ([]string, error) {
 	}
 
 	// Get the list of containers from the component
-	containerList, err = s3Component.ListBuckets()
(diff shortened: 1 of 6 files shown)


jfantinhardesty (approved):
LGTM!

PR #599, opened 2025-07-03
Log collector cmd

Dabnsky:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

This adds a gatherLogs command to the cloudfuse CLI that collects log files into an archive file

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #


cmd/log-collector.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+		defer dstFile.Close()
+
+		_, err = io.Copy(dstFile, srcFile)
+		if err != nil {
+			return err
+		}
+
+	}
+	return err
```
jfantinhardesty, line 370: This should just return nil, since it would have succeeded.

cmd/log-collector.go
```diff
@@ -0,0 +1,473 @@
...
+		if err != nil {
+			return err
+		}
+		if info.IsDir() {
+			if relPath == "." {
+				return nil
+			}
+			_, err := zipWriter.Create(relPath + "/")
+			if err != nil {
+				zipWriter.Close()
```
jfantinhardesty, line 402: Defer zipWriter.Close is called above, so it is not needed here since it will be called on the return.

cmd/log-collector.go
```diff
@@ -0,0 +1,473 @@
...
+func checkPath(outPath string) error {
+	var err error
+	if outPath == "" {
+		dumpPath, err = os.Getwd()
+		if err != nil {
+			return err
+		}
+	} else {
+		if !common.DirectoryExists(outPath) {
+			return err
```
jfantinhardesty, line 175: Err is nil at this point, so this will look like it succeeded. A custom error will need to be returned here.

jfantinhardesty (changes requested):
Great work on the log collector. Just a few small suggestions. Also, it looks like some of the new tests are not passing in CI/CD.

cmd/log-collector.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+		return nil
+	},
+}
+
+// checkPath makes sure the path for archive creation is valid.
+func checkPath(outPath string) error {
+	var err error
+	if outPath == "" {
+		dumpPath, err = os.Getwd()
```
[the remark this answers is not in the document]
Dabnsky: I can certainly change the function name. I can add to the function description to reflect that it sets the default when the string is blank. . Do you have a better place in mind for using os.Getwd() for the empty string?

cmd/log-collector.go
```diff
@@ -0,0 +1,473 @@
...
+	}
+	return err
+}
+
+func createWindowsArchive(archPath string) error {
+	outFile, err := os.Create(dumpPath + "/cloudfuse_logs.zip")
+	if err != nil {
+		return nil
+	}
+	err = outFile.Sync()
```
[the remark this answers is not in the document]
Dabnsky: This is an artifact from troubleshooting a race condition with the zip file. It's probably no longer needed.

jfantinhardesty (approved):

PR #606, opened 2025-07-10
Scheduling Feature 

Jeva2021:

- [x] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

Features:
- Added a Scheduling feature via Cron to schedule uploads based on the users selected times. 
- The schedule is parsed from the .YAML file, then inserted into a struct, and then creates Cron jobs to run on mount. 
- If no schedule is provided, there will be a default Cron job added which uploads every 30 seconds.

Other Changes:
- Flushfileinternals' logic was changed to only immediately upload changes to existing files in the cloud for consistency purposes, otherwise any additions will be added to scheduleOps map to later be uploaded according to provided schedule.
- Changed logic of UploadPendingFiles to call FlushfileInternal to upload files. 
- Added dependencies for Cron 
- Documented the scheduler in both base and advanced YAMl files.

### Checklist

- [x] Tested locally
- [x] Added new dependencies
- [x] Updated documentation
- [x] Added tests

### Related Issues

- Related Issue #
- Closes #


component/file_cache/scheduler.go
```diff
@@ -0,0 +1,276 @@
...
+	var cfg Config
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, err
+	}
+	if err := yaml.Unmarshal(data, &cfg); err != nil {
+		return nil, err
+	}
+	return cfg.Schedule, nil
+}
```
foodprocessor, line 41: Since this is part of the file cache component, you should use its existing config parsing capability.
That would mean putting your schedule config entry in the `FileCacheOptions` struct, and putting any config parsing code you need into the `Configure` function in file_cache.go. Then you shouldn't need this function.

component/file_cache/scheduler.go
```diff
@@ -0,0 +1,276 @@
...
+	var schedule WeeklySchedule
+	if !config.IsSet("schedule") {
+		log.Info(
+			"FileCache::SetupScheduler : No schedule configuration found(Using default schedule)",
+		)
+		schedule = WeeklySchedule{
+			"Default": UploadWindow{
+				CronExpr: "0 * * * * *",
+				Duration: "59s",
+			},
```
foodprocessor, line 54: Is there a simpler way to set the default schedule in a way that would work as an "always on" mode?

component/file_cache/scheduler.go
```diff
@@ -0,0 +1,276 @@
...
+				window.CronExpr = cronStr
+			}
+			if durStr, ok := rawWindow["duration"].(string); ok {
+				window.Duration = durStr
+			}
+			schedule[day] = window
+			log.Info("FileCache::SetupScheduler : Parsed schedule %s: cron=%s, duration=%s",
+				day, window.CronExpr, window.Duration)
+		}
+	}
```
foodprocessor, line 85: I think once you move the schedule into the file_cache config section (and the `FileCacheOptions` struct), this whole parsing block can be simplified and moved into the Config function.

component/file_cache/scheduler.go
```diff
@@ -0,0 +1,276 @@
...
+		c.AddFunc(config.CronExpr, func() {
+			startFunc()
+			log.Info("[%s] Starting upload at %s\n", currentDay, time.Now().Format(time.Kitchen))
+
+			// Create a context with timeout for the duration of the window
+			window, cancel := context.WithTimeout(context.Background(), durationParsed)
+			defer cancel()
+
+			ticker := time.NewTicker(30 * time.Second)
+			defer ticker.Stop()
```
foodprocessor, line 138: I think this is essentially a polling loop with a 30s wait. This design could delay uploads by up to 30 seconds when the upload schedule is active.

foodprocessor (changes requested):

component/file_cache/file_cache_test.go
```diff
@@ -1449,6 +1499,1216 @@ func (suite *fileCacheTestSuite) TestFlushFileErrorBadFd() {
...
+
+	suite.assert.True(uploaded, "File was not uploaded by the background servicePendingOps process")
+
+	// Verify the file data was correctly uploaded
+	suite.assert.FileExists(filepath.Join(suite.fake_storage_path, file))
+	d, _ := os.ReadFile(filepath.Join(suite.fake_storage_path, file))
+	suite.assert.Equal(data, d)
+}
+
+func (suite *fileCacheTestSuite) TestScheduleUploadsViaCronIntegration() {
```
jfantinhardesty, line 1701: This test follows a lot of the implementation details of scheduling uploads. However, the tests should test the overall behavior and not replicate the details used in the functions. I would refactor several of these functions to simplify the test.

jfantinhardesty (changes requested):

component/file_cache/file_cache.go
foodprocessor, line 1: I think logic needs to be added in a few more places to update the scheduler when files are deleted or renamed. And the createEmptyFile logic needs attention too.
I forgot about these for a while. Looking at the places where NextComponent() is called would be a great way to start thinking about what would be needed.

foodprocessor (dismissed):
Good progress!

common/types.go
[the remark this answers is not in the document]
Jeva2021: Deleted the extras left over from network-fault-tolerance.

component/file_cache/file_cache.go
```diff
@@ -78,6 +80,16 @@ type FileCache struct {
...
+
+	stopAsyncUpload chan struct{}
+
+	scheduler *cron.Cron
+
+	schedule WeeklySchedule
+
+	uploadNotifyCh chan struct{}
+
+	alwaysOn bool
```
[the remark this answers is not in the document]
Jeva2021: Deleted the whitespace.

component/file_cache/scheduler.go
```diff
@@ -0,0 +1,244 @@
...
+func (fc *FileCache) servicePendingOps() {
+	log.Info("FileCache::servicePendingOps : Servicing pending uploads")
+
+	// check if we're connected (keep this safety check)
+	if !fc.cloudConnected() {
+		log.Info(
+			"FileCache::servicePendingOps : Cloud storage not connected, skipping upload cycle",
+		)
+		return
+	}
```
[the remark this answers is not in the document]
Jeva2021: Dropped anything related to offline conditions.

component/file_cache/file_cache.go
```diff
@@ -1627,15 +1678,13 @@ func (fc *FileCache) FlushFile(options internal.FlushFileOptions) error {
 	return fc.flushFileInternal(options)
 }
 
-// file must be locked before calling this function
```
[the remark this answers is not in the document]
jfantinhardesty: It looks like they aren't quite back yet. I would suggest looking at the diff in GitHub and it should show you the comments that are missing in the code.

jfantinhardesty (changes requested):
The tests look great! You did a great job planning out the different cases and tests.

component/file_cache/file_cache.go
Jeva2021: Added logic to rename and delete for scheduledOps.

jfantinhardesty (approved):

component/file_cache/scheduler.go
```diff
@@ -0,0 +1,261 @@
...
+					windowCount := fc.activeWindows
+					fc.activeWindowsMutex.Unlock()
+
+					log.Info("[%s] Upload window ended at %s (remaining windows: %d)\n",
+						windowName, time.Now().Format(time.Kitchen), windowCount)
+
+					// Only close resources when the last window ends
+					if isLastWindow {
+						close(fc.uploadNotifyCh)
+						fc.uploadNotifyCh = nil
```
foodprocessor, line 126: As a pattern, it's dangerous to close a channel on a different thread than senders write to that channel.
Setting the channel to nil may improve the situation, but it doesn't fix the issue - a sender might try to send on the closed channel due to a race condition, and then the whole process will panic and crash.
Also setting the channel pointer to nil may cause a race condition with the channel's receivers.
The best way to eliminate this issue is to create another channel to tell the receiver that the schedule window is closing, and then leave this channel open and let it sit with a signal in it while the schedule is inactive.
If we allowed messages to pile up in the buffered channel, this would be a problem, but that's where the `select`, `case myChan <- struct{}{}` ... `default` patter comes in. Then we can write to the channel without blocking, and messages don't pile up. The channel just tells the receiver if there is anything to upload, since there is no need to tell it how many uploads there are to do.

PR #633, opened 2025-08-08
Add sync command

jfantinhardesty:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Add sync command for consistency with size tracker. New sync command will calculate the size of a subdirectory and then change the size_tracker file to account for the new size.

### Checklist

- [ ] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/cmd/sync.go
+++ b/cmd/sync.go
@@ -1,0 +1,169 @@
+/*
+   Licensed under the MIT License <http://opensource.org/licenses/MIT>.
+
+   Copyright © 2023-2025 Seagate Technology LLC and/or its Affiliates
+
+   Permission is hereby granted, free of charge, to any person obtaining a copy
+   of this software and associated documentation files (the "Software"), to deal
+   in the Software without restriction, including without limitation the rights
+   to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+   copies of the Software, and to permit persons to whom the Software is
+   furnished to do so, subject to the following conditions:
+
+   The above copyright notice and this permission notice shall be included in all
+   copies or substantial portions of the Software.
+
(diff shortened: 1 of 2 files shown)


cmd/sync.go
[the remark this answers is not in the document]
jfantinhardesty: I think not, I think this should really only be used with the plugin.

component/size_tracker/size_tracker.go
```diff
@@ -106,14 +105,14 @@ func (st *SizeTracker) Configure(_ bool) error {
...
 				journalName = sanitizedName + ".dat"
 			}
 		} else {
 			azconf := azstorage.AzStorageOptions{}
 			if err := config.UnmarshalKey("azstorage", &azconf); err == nil {
-				sanitizedName := sanitizeFileName(azconf.Container + "-" + azconf.PrefixPath)
+				sanitizedName := common.SanitizeName(azconf.Container + "-" + azconf.PrefixPath)
 				if sanitizedName != "" {
 					journalName = sanitizedName + ".dat"
 				}
```
foodprocessor, line 118: Isn't this feature S3-specific?

foodprocessor (approved):
I get the nagging feeling that we should move the S3 code into the s3storage component and add a new component interface function, but the feeling is not strong enough to hold this up.

component/size_tracker/size_tracker.go
jfantinhardesty: Yes, yes it is. I was just using what I had already written which I think was being overly ambitious that we should support both. I can remove this extra Azure code if you want.

PR #641, opened 2025-08-14
Add TUI 

brayan-trejo:

- [ ] Refactor
- [X ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

- Added /cmd/config.go (source code, called by using `cloudfuse config`)
- Creates **_config.aes_** in project root. 

FUTURE CHANGES: 
- Add dry run mount call to make sure configuration was successful before letting the user go. 
- Add credential key validation. There are some built-in validation when calling `getContainerListAzure()` or `getContainerListS3()` but need to do further testing. 
- Add a check that makes sure the available cache size is at least XGB when caching is enabled. Otherwise, recommend that user disables caching. Where X is the minimum required capacity for caching. 

### Checklist

- [X ] Tested locally
- [X ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


cmd/config.go
```diff
@@ -0,0 +1,23 @@
+package cmd
```
jfantinhardesty, line 1: The license should be at the top of this file.

cmd/tui.go
```diff
@@ -0,0 +1,1432 @@
...
+	navigationHomeLabel string = 	"[black]🏠 Home[-]"
+	navigationNextLabel string = 	"[black]🡲  Next[-]"
+	navigationBackLabel string = 	"[black]🡰  Back[-]"
+	navigationPreviewLabel string = "[black]📄 Preview[-]"
+	navigationQuitLabel string = 	"[black]❌ Quit[-]"
+	navigationFinishLabel string = 	"[black]✅ Finish[-]"
+	navigationWidgetHeight int = 3
+)
+
+type Config struct {
```
jfantinhardesty, line 94: These types don't need to start with capital letters (Config, LibfuseConfig, etc.). Starting with a capital means they are exported and none of these need to be used outside of this file.

cmd/tui.go
```diff
@@ -0,0 +1,1432 @@
...
+	It returns the full path to the cache directory.
+*/
+func getDefaultCachePath() string {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		// TODO: Handle error if home directory cannot be retrieved.
+		// TODO: Choose a fallback path for cache directory.
+		return "/tmp/cloudfuse/cloudfuse.log"
+	}
+	filepath := filepath.Join(home, ".cloudfuse", "file_cache")
```
jfantinhardesty, line 1184: I would rename the filepath variable. It shadows the filepath package and could cause issues later if you need to call filepath.Join again.

cmd/tui.go
```diff
@@ -0,0 +1,1432 @@
...
+	Type 			 string `yaml:"type"`
+    AccountName      string `yaml:"account-name"`
+    AccountKey       string `yaml:"account-key"`
+    Endpoint         string `yaml:"endpoint,omitempty"`
+    Mode             string `yaml:"mode,omitempty"`
+    Container  		 string `yaml:"container"`
+}
+
+
+/*
```
jfantinhardesty, line 137: Could you change the multiline comments above each function to be regular comments with slashes. This would better follow the rest of the style of comments in other files.

jfantinhardesty (changes requested):
Looks good, just a few issues to resolve. Love the look of the TUI!

cmd/tui.go
```diff
@@ -0,0 +1,1376 @@
...
+
+type libfuseConfig struct {
+    NetworkShare bool `yaml:"network-share"`
+}
+
+type attrCacheConfig struct {
+    TimeoutSec int `yaml:"timeout-sec"`
+}
+
+type fileCacheConfig struct {
```
jfantinhardesty, line 108: One thought, could you use FileCacheOptions from the file_cache.go file here instead? And similar ones for the other config options. This way we don't duplicate the names in the yaml config in case they change later.

brayan-trejo: Great suggestion. Latest commit changes to using existing options structs with the excpetion of S3 since there is bit more work to do with the security enclave handling of the keys.

cmd/tui.go
```diff
@@ -0,0 +1,1405 @@
...
+	enableCaching             bool         = true                    // If true, sets cacheMode to file_cache. If false, block_cache
+	cacheLocation             string       = getDefaultCachePath()   // Sets file_cache.path @ startup to default: $HOME/.cloudfuse/cache
+	cacheSize                 string       = "80"                    // User-defined cache size as %
+	availableCacheSizeGB      int                                    // Total available cache size in GB @ the cache location
+	currentCacheSizeGB        int                                    // Current cache size in GB based on 'cacheSize' percentage
+	clearCacheOnStart         bool         = false                   // If false, sets 'allow-non-empty-temp' to true
+	cacheRetentionDuration    int          = 2                       // User-defined cache retention duration. Default is '2'
+	cacheRetentionUnit        string                                 // User-defined cache retention unit (sec, min, hours, days). Default is 'days'
+	cacheRetentionDurationSec int                                    // Sets 'file_cache.timeout-sec' from 'cacheRetentionDuration'
```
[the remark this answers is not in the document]
brayan-trejo: Fixed. Created new structs and refactored code to adopt OOP design.

cmd/tui.go
```diff
@@ -0,0 +1,1415 @@
...
+		return fmt.Errorf("failed to marshal YAML: %v", err)
+	}
+
+	tmpFile := "config-tmp.yaml"
+	if err := os.WriteFile(tmpFile, yamlData, 0600); err != nil {
+		return fmt.Errorf("failed to write YAML to file: %v", err)
+	}
+
+	// Update options.ConfigFile to point to the temporary file
+	options.ConfigFile = "config-tmp.yaml"
```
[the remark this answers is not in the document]
brayan-trejo: Instead of writing to disk, I removed createTmpConfigFile() and just read from buffer instead to avoid ever writing keys to disk.

cmd/tui.go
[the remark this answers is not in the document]
brayan-trejo: Done!

cmd/config.go
```diff
@@ -0,0 +1,1496 @@
...
+
+func init() {
+	rootCmd.AddCommand(configCmd)
+}
+
+// Main function to run the TUI application.
+// Initializes the tview application, builds the TUI application, and runs it.
+func (tui *appContext) runTUI() error {
+	// Disable cloudfuse logging during TUI session to prevent log messages from interfering with the UI.
+	log.SetLogLevel(1)
```
jfantinhardesty, line 180: Good catch noticing the logs were causing issues with the TUI. I still see that one line prints to the terminal. You could change this to use silent logging, which basically disables logging by doing something like log.SetDefaultLogger("silent", common.LogConfig{Level: common.ELogLevel.LOG_DEBUG()})

brayan-trejo: Worked perfectly. Fixed. Thanks!

cmd/config.go
```diff
@@ -0,0 +1,1496 @@
...
+	// Run the application
+	if err := tui.app.Run(); err != nil {
+		panic(err)
+	}
+
+	return nil
+}
+
+// Function to build the TUI application. Initializes the pages and adds them to the page stack.
+func (tui *appContext) buildTUI() {
```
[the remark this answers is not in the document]
brayan-trejo: I agree. looks cleaner! Done

PR #705, opened 2025-09-23
Network fault tolerance

foodprocessor:
This adds an offline access feature in the file cache.
This feature is enabled by default.


common/types.go
```diff
@@ -123,6 +123,28 @@ func (e CloudUnreachableError) Is(target error) bool {
 	return ok
 }
 
+type NoCachedDataError struct {
+	Message    string
+	CacheError error
+}
+
+func NewNoCachedDataError(originalError error) CloudUnreachableError {
+	return CloudUnreachableError{
```
jfantinhardesty: Shouldn't this return a NoCachedDataError

component/file_cache/async.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+// checks whether we have usable metadata, despite being offline
+func offlineDataAvailable(err error) bool {
+	return isOffline(err) && cachedData(err)
+}
+
+// checks whether we have usable metadata, despite being offline
+func cachedData(err error) bool {
+	return !errors.Is(err, &common.NoCachedDataError{}) || !isOffline(err)
```
jfantinhardesty: Could you change this to !(errors.Is(err, &common.NoCachedDataError{}) && isOffline(err)). This seems more readable to me.

component/file_cache/file_cache.go
```diff
@@ -555,22 +537,121 @@ func isLocalDirEmpty(path string) bool {
...
+		// block directory creation (to protect data consistency)
+		log.Warn(
+			"FileCache::CreateDir : %s might exist in cloud storage. Creation is blocked.",
+			options.Name,
+		)
+	default:
+		// the directory already exists in cloud storage
+		err = os.ErrExist
+		// use distinct log messages for when the attribute cache entry is valid or expired
+		if err == nil { // valid
```
jfantinhardesty, line 610: Isn't this case impossible. err was set to os.ErrExist.

component/file_cache/OfflineAccess.md
```diff
@@ -0,0 +1,56 @@
+# Offline Access
```
jfantinhardesty, line 1: We should also note that if cloudfuse is shutdown, that pending uploads will be lost.

component/file_cache/async.go
```diff
@@ -0,0 +1,379 @@
...
+func (fc *FileCache) servicePendingOps() {
+	for {
+		select {
+		case <-fc.componentStopping:
+			log.Crit("FileCache::servicePendingOps : Stopping")
+			// TODO: Persist pending ops
+			return
+		case <-fc.startScheduledUploads:
+			// check if we're connected
+			if !fc.NextComponent().CloudConnected() {
```
jfantinhardesty, line 230: Should this check use exponential backoff rather than a 1 second loop?

component/file_cache/file_cache.go
```diff
@@ -803,8 +931,21 @@ func (fc *FileCache) RenameDir(options internal.RenameDirOptions) error {
 
 	// rename the directory in the cloud
 	err = fc.NextComponent().RenameDir(options)
-	if err != nil {
-		log.Err("FileCache::RenameDir : error %s [%s]", options.Src, err.Error())
+	// if we are offline and offline access is enabled, allow fully cached local directories to be renamed
```
jfantinhardesty, line 936: Is this an issue if we only have a limited listing of a directory, or an outdated one? We would rename the entries we have, but others wouldn't right. I thinm that is probably fine, but should be documented in the markdown file.

component/file_cache/file_cache.go
```diff
@@ -691,8 +786,39 @@ func (fc *FileCache) IsDirEmpty(options internal.IsDirEmptyOptions) bool {
...
+	attrs, _, err := fc.NextComponent().StreamDir(sdOptions)
+	if !cachedData(err) {
+		log.Err("FileCache::IsDirEmpty : %s no attribute cache data (offline)", options.Name)
+		return false
+	}
+	if len(attrs) > 0 {
+		log.Debug("FileCache::IsDirEmpty : %s not empty in attribute cache (offline)", options.Name)
+		return false
+	}
+	// attribute cache has no entries for this directory, but we need to check if it has a complete listing
```
jfantinhardesty, line 811: This feels a bit janky. Is there a better way to check for the listing complete flag?

component/file_cache/lru_policy.go
```diff
@@ -77,17 +77,21 @@ type lruPolicy struct {
...
 }
 
-// LRUPolicySnapshot represents the *persisted state* of lruPolicy.
+// lruPolicySnapshot represents the *persisted state* of lruPolicy.
 // It contains only the fields that need to be saved, and they are exported.
-type LRUPolicySnapshot struct {
-	NodeList           []string // Just node names, *without their fc.tmp prefix*, in linked list order
-	CurrMarkerPosition uint64   // Node index of currMarker
-	LastMarkerPosition uint64   // Node index of lastMarker
+type lruPolicySnapshot struct {
```
jfantinhardesty, line 89: Why the change to lower case for the LRUPolicySnapshot?

common/types.go
foodprocessor: Yes!

component/file_cache/OfflineAccess.md
foodprocessor: They're actually persisted on shutdown (see component/file_cache/lru_policy.go:203).

component/file_cache/async.go
foodprocessor: This one second loop calls CloudConnected, which serves a cached response until the exponential backoff triggers a new check. Should I add a comment here to that effect?

component/file_cache/file_cache.go
foodprocessor: Woops! Good catch! I moved this down.

component/file_cache/file_cache.go
foodprocessor: It is janky! But this is the price of not changing the interface! 😛

component/file_cache/OfflineAccess.md
jfantinhardesty: 😲

component/file_cache/file_cache.go
foodprocessor: You mean if we *had* a complete listing, but the cloud contents changed, and we did an offline rename?
Yes, that would make for a mess. There are warnings in the documentation about not making parallel changes to the same objects or prefixes. I can make sure it's as clear as it can be.

component/file_cache/lru_policy.go
foodprocessor: I originally made this upper case because I was confused and thought it was a gob requirement.

component/file_cache/async.go
jfantinhardesty: Yeah a comment might be good, just because it looks like we are doing a costly polling loop, when the backoff is elsewhere.

jfantinhardesty (approved):

PR #706, opened 2025-09-26
Cleanup documentation a bit

foodprocessor:
Reconciling differences between baseConfig.yaml and advancedConfig.yaml.


jfantinhardesty (approved):
Thank you for cleaning up these docs!

PR #707, opened 2025-09-26
Change Nx size target code to work with intermittent size updates

foodprocessor:
Capture the factor difference between this mount size and the bucket size, whenever the bucket size is updated.
Use that factor to set a target size for the mount, and translate that into an offset to use until the next bucket size update.

### DIFF
--- a/component/size_tracker/size_tracker.go
+++ b/component/size_tracker/size_tracker.go
@@ -302,23 +302,23 @@ func (st *SizeTracker) StatFs() (*common.Statfs_t, bool, error) {
 		stat, ret, err := st.NextComponent().StatFs()
 
 		if err == nil && ret {
-			bucketUsage_uint64 := stat.Blocks * uint64(blockSize)
+			returnedBucketUsage := stat.Blocks * uint64(blockSize)
+			// convert everything to float64
+			bucketCapacity := float64(st.totalBucketCapacity)
+			bucketUsage := float64(returnedBucketUsage)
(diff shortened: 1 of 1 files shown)


component/size_tracker/size_tracker.go
```diff
@@ -265,19 +276,29 @@ func (st *SizeTracker) StatFs() (*common.Statfs_t, bool, error) {
...
+			// Check if cloud size has changed
+			bucketSize := stat.Blocks * uint64(blockSize)
+			if bucketSize != st.bucketSize && bucketSize > 0 {
+				// since the bucket size has been updated, assume it's a recent update
+				// so now we can correlate the relative size of this mount to the entire bucket
+
+				// the target is to fill the entire bucket to the eviction threshold
+				targetBucketSize := evictionThreshold * float64(st.totalBucketCapacity)
+				// what factor is that, from where we are now?
+				targetFactor := targetBucketSize / float64(bucketSize)
```
jfantinhardesty, line 288: I think there are some issues with the targetFactor here. It will be very large when there is little data written. So when that is multiplied by the targetSize it will be significantly larger than the evictionTriggerSize, so the statSizeOffset will be negative which converted to a uint64 will be very large.

foodprocessor: Good point. We never want a negative offset.
I went ahead and fixed that with a call to `min`.
Now the target size is not allowed to exceed the eviction threshold based on the display capacity.

PR #831, opened 2026-02-12
fix: resolve cron scheduler goroutine leak and modernize usage

CybotTM:
## Summary

- **Fix goroutine leak**: store `*cron.Cron` instance and call `Stop()` during shutdown
- **Replace hand-rolled validation**: use go-cron's built-in `ValidateSpec()` instead of custom `isValidCronExpression()`
- **Upgrade go-cron** to [v0.11.0](https://github.com/netresearch/go-cron/releases/tag/v0.11.0)

## Context

Found during [netresearch/go-cron dependent analysis](https://github.com/netresearch/go-cron/releases/tag/v0.11.0). The cron scheduler was created in `SetupScheduler()` but never stored as a struct field, making it impossible to stop — causing a goroutine leak for the lifetime of the process.

## Changes

| File | Change |
|------|--------|
| `component/azstorage/azautolog.go` | Store `*cron.Cron` as struct field, call `Stop()` in `Stop()` method |
| `component/azstorage/azautolog.go` | Replace `isValidCronExpression()` with `cron.ValidateSpec()` |
| `go.mod` / `go.sum` | Upgrade go-cron to v0.11.0 |

## Test plan

- [ ] Verify `Stop()` is called during component shutdown
- [ ] Verify cron expression validation still works
- [ ] Verify scheduled uploads still fire correctly

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -45,6 +45,7 @@ import (
 	"github.com/Seagate/cloudfuse/internal"
 	"github.com/Seagate/cloudfuse/internal/handlemap"
 	"github.com/Seagate/cloudfuse/internal/stats_manager"
+	cron "github.com/netresearch/go-cron"
 )
 
 // Common structure for Component
@@ -82,6 +83,7 @@ type FileCache struct {
 	activeWindows      int
 	activeWindowsMutex *sync.Mutex
 	closeWindowCh      chan struct{}
+	cronScheduler      *cron.Cron
 }
 
 // Structure defining your config parameters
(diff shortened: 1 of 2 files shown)


foodprocessor:
Thank you for your contribution!
We're working to wrap up a new feature that touches this same code, so I'm thinking to merge your contribution into that feature branch before it merges to main.
Thank you for your patience!

foodprocessor (approved):
Good changes all around. Thank you!

foodprocessor:
I went ahead and merged this directly into main.
Thanks again for your contribution and your patience!

PR #897, opened 2026-04-22
Always flush on sync (remove flags)

foodprocessor:
Sync now always just calls flush, but with the CloseInProgress flag, which ensures flush actually happens and is not put off (in case of LazyWrite).
This is primarily a readability improvement, and does not change default functionality.

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -65,7 +65,6 @@ type FileCache struct {
 	allowOther      bool
 	offloadIO       bool
 	syncToFlush     bool
-	syncToDelete    bool
 	maxCacheSizeMB  float64
 
 	defaultPermission os.FileMode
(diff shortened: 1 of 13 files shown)


component/file_cache/file_cache.go
```diff
@@ -1689,25 +1664,7 @@ func (fc *FileCache) WriteFile(options *internal.WriteFileOptions) (int, error)
...
-		if err != nil {
-			log.Err("FileCache::SyncFile : %s failed", options.Handle.Path)
-			return err
-		}
-
-		options.Handle.Flags.Set(handlemap.HandleFlagFSynced)
-	}
-
-	return nil
+	return fc.FlushFile(internal.FlushFileOptions{Handle: options.Handle, CloseInProgress: true})
```
jfantinhardesty: Do we care here that we don't call NextComponent SyncFile? I know the original code didn't but curious what SyncFile is used for by the other components?

jfantinhardesty (approved):

component/file_cache/file_cache.go
foodprocessor: Good question, I forgot about that call!
Below the file cache, only the attribute cache implements it. It invalidates the file's attributes.
I'll add that back in. Thanks!

PR #909, opened 2026-05-06
Add listingComplete flag

foodprocessor:
This new flag tracks whether we have a complete listing of all the contents in a directory. This makes it easy to serve ENOENT from the attribute cache without wasting cloud calls.
Also included attribute cache improvements for performance, logic, and removing dead code.

### DIFF
--- a/component/attr_cache/attr_cache.go
+++ b/component/attr_cache/attr_cache.go
@@ -1285,9 +1285,7 @@ func (ac *AttrCache) FlushFile(options internal.FlushFileOptions) error {
 		toBeInvalid, found := ac.cache.get(options.Handle.Path)
 		if found {
 			toBeInvalid.invalidate()
-		}
-		if parent, found := ac.cache.get(getParentDir(options.Handle.Path)); found &&
-			parent.exists() {
+		} else if parent, found := ac.cache.get(getParentDir(options.Handle.Path)); found && parent.exists() {
 			parent.listCache = nil
 			parent.listingComplete = false
 		}
@@ -1340,9 +1338,7 @@ func (ac *AttrCache) CommitData(options internal.CommitDataOptions) error {
 		entry, found := ac.cache.get(options.Name)
 		if found {
 			entry.invalidate()
-		}
-		// make sure parent gets a fresh listing after this change
-		if parent, found := ac.cache.get(getParentDir(options.Name)); found && parent.exists() {
+		} else if parent, found := ac.cache.get(getParentDir(options.Name)); found && parent.exists() {
 			parent.listCache = nil
 			parent.listingComplete = false
 		}
--- a/component/attr_cache/attr_cache.go
+++ b/component/attr_cache/attr_cache.go
@@ -478,12 +478,12 @@ func (ac *AttrCache) CreateDir(options internal.CreateDirOptions) error {
 			})
 			// insert returns nil when entries are maxed out
 			if dirAttrCacheItem != nil {
-				// update flag for tracking directory existence
 				if ac.cacheDirs {
+					// update flag for tracking directory existence
 					dirAttrCacheItem.markInCloud(false)
+					// this is a new directory, so we have a complete (empty) listing for it
+					dirAttrCacheItem.listingComplete = true
 				}
-				// this is a new directory, so we have a complete (empty) listing for it
-				dirAttrCacheItem.listingComplete = true
 			}
 			// if this is a new entry, update the parent directory timestamps
 			if err == nil {
--- a/component/attr_cache/attr_cache.go
+++ b/component/attr_cache/attr_cache.go
@@ -750,8 +750,17 @@ func (ac *AttrCache) IsDirEmpty(options internal.IsDirEmptyOptions) bool {
 	}
 	// Is the directory in our cache?
 	ac.cacheLock.RLock()
-	defer ac.cacheLock.RUnlock()
 	item, found := ac.cache.get(options.Name)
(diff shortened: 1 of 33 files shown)


jfantinhardesty (changes requested):
For the listing complete flag, you should expire that flag after the expiration time has elapsed.

brayan-trejo (approved):

component/attr_cache/attr_cache.go
```diff
@@ -737,14 +745,15 @@ func (ac *AttrCache) IsDirEmpty(options internal.IsDirEmptyOptions) bool {
...
 			options.Name,
 		)
+		// when offline, this will return false
 		return ac.NextComponent().IsDirEmpty(options)
 	}
 	// Is the directory in our cache?
 	ac.cacheLock.RLock()
-	pathInCache := ac.pathExistsInCache(options.Name)
-	ac.cacheLock.RUnlock()
+	defer ac.cacheLock.RUnlock()
```
jfantinhardesty: I don't think you need to defer this unlock. We previously unlocked it right after we check the cache. Are there operations after this that still need the lock?

component/attr_cache/attr_cache.go
```diff
@@ -765,16 +779,10 @@ func (ac *AttrCache) IsDirEmpty(options internal.IsDirEmptyOptions) bool {
...
-	ac.cacheLock.RLock()
-	defer ac.cacheLock.RUnlock()
-
-	directory, found := ac.cache.get(prefix)
-	if found && directory.exists() {
-		for _, chldItem := range directory.children {
-			if chldItem.exists() {
-				return true
-			}
+func (value *attrCacheItem) hasExistingChildren() bool {
```
jfantinhardesty: I kind of prefer the original code which acquired the lock inside the function. Otherwise we need to know that you need to be holding the lock to call this function.

component/attr_cache/attr_cache.go
```diff
@@ -1157,28 +1169,53 @@ func (ac *AttrCache) GetAttr(options internal.GetAttrOptions) (*internal.ObjAttr
...
+	if err != nil && !os.IsNotExist(err) {
+		return pathAttr, err
+	}
+	// response is valid - cache it
 	ac.cacheLock.Lock()
 	defer ac.cacheLock.Unlock()
-
-	switch err {
-	case nil:
+	if err == nil {
```
jfantinhardesty: Any reason changing the switch. I think this was previously auto applied from a go cleanup script.

component/attr_cache/attr_cache.go
```diff
@@ -466,33 +456,43 @@ func (ac *AttrCache) CreateDir(options internal.CreateDirOptions) error {
...
+				cachedAt: currentTime,
+			})
+			// insert returns nil when entries are maxed out
+			if dirAttrCacheItem != nil {
+				// update flag for tracking directory existence
+				if ac.cacheDirs {
+					dirAttrCacheItem.markInCloud(false)
+				}
+				// this is a new directory, so we have a complete (empty) listing for it
+				dirAttrCacheItem.listingComplete = true
```
jfantinhardesty: This runs in the case that the next component returns EEXIST. Is that a problem?

component/attr_cache/attr_cache.go
```diff
@@ -727,6 +725,16 @@ func (ac *AttrCache) cacheListSegment(
 		listDirPath, token, nextToken, len(pathList))
 }
 
+func (ac *AttrCache) markListingComplete(listDirPath string) {
+	ac.cacheLock.Lock()
+	defer ac.cacheLock.Unlock()
+	listDirItem, found := ac.cache.get(listDirPath)
+	if found {
+		listDirItem.listingComplete = true
+		listDirItem.cachedAt = time.Now()
```
jfantinhardesty, line 733: It seems that cachedAt is being asked to do a lot, and her the time gets updated. Should there be a separate timestamp for the listingComplete flag, rather than using cachedAt?

foodprocessor: I originally wrote it with a separate timestamp, but I noticed that there was no conflict between the two, and it seemed like a waste to add a timestamp value we don't need. The semantics of the two timestamps is nearly identical (when we last checked this directory's attributes in cloud storage vs. when we last listed this directory), especially since directories have limited metadata (their size doesn't change, and the timestamp is managed more on the local side than in cloud storage).
So, it seemed like a good fit.
What do you think?

component/attr_cache/attr_cache.go
foodprocessor: Good catch! I'll move this into the cacheDirs block.

component/attr_cache/attr_cache.go
foodprocessor: I think the lock should be held continuously, from the `get` call here, to the `hasExistingChildren` call below, to avoid a race condition.
I think you're right though, that defer is not the right solution, because it could end up locked during a network call.
I'll rewrite it to shorten the lock around the critical section.

component/attr_cache/attr_cache.go
foodprocessor: At this point, this function is only called once. I'll go ahead and reintegrate it into its one caller.

component/attr_cache/attr_cache.go
foodprocessor: I don't know why I changed this... and it's a bad change, to boot.
I'm reverting this.

jfantinhardesty (approved):

PR #910, opened 2026-05-06
Add context and connection state to cloud storage components

foodprocessor:
Add CloudConnected component function, and make cloud storage components stateful (connection-aware).
This is needed for the network fault tolerance feature.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -100,6 +100,29 @@ func CloudfuseVersion_() string {
 	return cloudfuseVersion_
 }
 
+// custom errors shared by different components
+type CloudUnreachableError struct {
+	Message           string
+	CloudStorageError error
+}
+
+func NewCloudUnreachableError(originalError error) CloudUnreachableError {
+	return CloudUnreachableError{
+		Message:           "Failed to connect to cloud storage",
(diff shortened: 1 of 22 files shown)


jfantinhardesty (changes requested):
Can you change the timeout to instead of always checking every ten seconds if the cloud is up, to move this to CreateDir and have a similar timeout there since that is the only case where we need to know if the cloud is down.

foodprocessor:
> Can you change the timeout to instead of always checking every ten seconds if the cloud is up, to move this to CreateDir and have a similar timeout there since that is the only case where we need to know if the cloud is down.

I described this incorrectly (because I forgot how it works!).
The health check ticker actually only runs when it goes offline.
Then it attempts `HeadBucket`, with exponential backoff (maxing out at 90s), until it gets back online.
When it's connected, even pipeline calls to `CloudConnected()` don't make network calls. It's just a way for the cloud component to share its connection state up the pipeline.
So this feature only adds a single cloud storage request *per outage*.

jfantinhardesty (approved):

brayan-trejo (approved):

PR #911, opened 2026-05-06
Scrub S3

foodprocessor:
Change DeleteDirectory to only delete a marker, especially since it is only called on empty directories.
Optimize GetAttr performance (don't call List when directory markers may be present).


jfantinhardesty (approved):
Nice change. Thought for later is would we ever want a flag that enables deleting full directories, since it is very slow to do large deletion with the current code. Then that original code could be used.

foodprocessor:
> Nice change. Thought for later is would we ever want a flag that enables deleting full directories, since it is very slow to do large deletion with the current code. Then that original code could be used.

At the moment, we use a test script for bulk deletion, and we recommend users use a separate tool to clear out their buckets. We could expose a separate cleanup function to do bulk deletion in the future, but I'd wait for someone to ask for it.

PR #913, opened 2026-05-07
Prevent async validate calls from resurrecting deleted nodes

foodprocessor:


component/file_cache/lru_policy.go
```diff
@@ -351,27 +351,34 @@ func (p *lruPolicy) asyncCacheValid() {
 		case name := <-p.validateChan:
 			// validateChan only gets names that are already cached
 			// if the file is not in the map anymore, then it was deleted,
-			// which means calling cacheValidate now would be a bug
-			_, found := p.nodeMap.Load(name)
-			if found {
-				p.cacheValidate(name)
-			}
+			// so call cacheValidate with create=false (don't resurrect deleted nodes)
+			p.cacheValidate(name, false)
```
jfantinhardesty, line 355: This change seems overly complicated. Wouldn't the change just be to remove the if found {} case?

jfantinhardesty (approved):

brayan-trejo (approved):

PR #922, opened 2026-05-18
Cancelled context errors don't update connection state

foodprocessor:
A minor change to prevent us from treating our offline state as new information.

The purpose of this change is to avoid a race condition, where the network comes back, and two things happen at once:

1. The component checks the cloud and sees it's actually connected.
2. Another cloud call comes in from a host application and hits the cancelled context.

Then it would be possible for there to be two connection state updates calling `updateConnectionState` simultaneously. If the retry mechanism gets in first, it sets `firstOffline` to nil (we're connected), then the failed host app call updates the connection state next, and falsely identifies the cancelled context as a new connection failure - bring the whole component back offline until the next retry.

I see this change as the simplest way to eliminate this problem. Just don't treat a cancelled context as *evidence* of a failure. Because it isn't. It's just a *result* of a failure.

### DIFF
--- a/component/azstorage/azstorage.go
+++ b/component/azstorage/azstorage.go
@@ -264,6 +264,13 @@ func (az *AzStorage) timeToRetry() bool {
 func (az *AzStorage) updateConnectionState(err error) bool {
 	az.state.Lock()
 	defer az.state.Unlock()
+
+	// A context.Canceled error means az.ctx was already cancelled by us when we went offline.
+	// It carries no new connectivity information, so we must not update state.
+	if errors.Is(err, context.Canceled) {
+		return az.state.firstOffline == nil
+	}
+
 	currentTime := time.Now()
 	az.state.lastConnectionAttempt = &currentTime
 	connected := !isOfflineError(err)
--- a/component/s3storage/s3storage.go
+++ b/component/s3storage/s3storage.go
@@ -59,7 +59,6 @@ type connectionState struct {
 	lastConnectionAttempt *time.Time
 	firstOffline          *time.Time
 	retryTicker           *time.Ticker
-	contextEpoch          uint64
 }
 
 const compName = "s3storage"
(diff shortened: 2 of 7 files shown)


component/azstorage/azstorage.go
```diff
@@ -264,6 +264,13 @@ func (az *AzStorage) timeToRetry() bool {
 func (az *AzStorage) updateConnectionState(err error) bool {
 	az.state.Lock()
 	defer az.state.Unlock()
+
+	// A context.Canceled error means az.ctx was already cancelled by us when we went offline.
+	// It carries no new connectivity information, so we must not update state.
+	if errors.Is(err, context.Canceled) {
```
brayan-trejo, line 270: Is there ever an instance where `context.Canceled` occurs from something else other than going offline? If so, it might be good to change this conditional to ignore state updates only when we are already offline:

```
if errors.Is(err, context.Canceled) && az.state.firstOffline != nil {
    return false
}
```

This way, if we still see the `context.Canceled` while we're still online that, that means we're something is wrong and we don't ignore it

component/s3storage/s3storage.go
```diff
@@ -237,6 +237,13 @@ func (s3 *S3Storage) timeToRetry() bool {
 func (s3 *S3Storage) updateConnectionState(err error) bool {
 	s3.state.Lock()
 	defer s3.state.Unlock()
+
+	// A context.Canceled error means s3.ctx was already cancelled by us when we went offline.
+	// It carries no new connectivity information, so we must not update state.
+	if errors.Is(err, context.Canceled) {
```
brayan-trejo, line 243: Same thing here as with the az component

brayan-trejo (approved):
Looks good! My two comments stem from not knowing exactly what can cause the `context.Canceled` which you probably already considered.

component/azstorage/azstorage.go
foodprocessor: I'm realizing I should have given this PR a proper description. I just updated that, and it should clear this up a bit.
My understanding is that cloud network calls won't return a cancelled context error on their own.

foodprocessor:
I've added several new unit tests to more thoroughly test the offline access feature.

foodprocessor:
PendingOps entries are now removed when a file is uploaded by ReleaseFile or FlushFile successfully (prevent double-uploads in this corner case).
I've also added LocalStack to the block and connect test scripts.

PR #928, opened 2026-05-23
Fix deep CI bug due to eventual consistency

foodprocessor:
A recurring CI failure uncovered a minor inconsistency in our directory listing with eventually consistent cloud storage:
If we list a directory *right* after creating a file or directory, we cache the new entity's attributes in the attribute cache, but the listing may not include it. This caused the entry to be excluded entirely because it would fall between the cracks - neither truly local-only, nor included in the cloud listing.
This change mitigates the issue by at least including the entry if it's in the attribute cache, and it clearly should have been in the current page of the listing.

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -721,15 +721,52 @@ func (fc *FileCache) StreamDir(
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


jfantinhardesty (approved):
LGTM!

PR #942, opened 2026-06-03
Fix gosec security issues

jfantinhardesty:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

Add gosec as a linter which finds go security issues and fixed identified issues.

### Checklist

- [ ] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,6 +6,7 @@ linters:
   enable:
     - errcheck # Errcheck is a program for checking for unchecked errors in go programs.
     - gocheckcompilerdirectives # Checks that go compiler directive comments (//go:) are valid.
+    - gosec # Inspects source code for security problems.
     - govet # Vet examines Go source code and reports suspicious constructs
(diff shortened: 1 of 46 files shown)


foodprocessor (changes requested):
I don't think we should check for overflow in cases where it will never happen. To fix the vulnerability, we can and should normalize incoming values to make overflow impossible. I'm grateful to gosec for highlighting the issue, but I don't think the gosec integer overflow checker is mature enough to be included in our CI.
So, I think we should add G115 to the excludes list, and this PR should pivot to addressing the problem in the best way without affecting readability (normalize incoming values and change variable types to remove the need for conversions, where possible).

jfantinhardesty:
Closing due to new PR which addresses everything but the G115 check.

PR #960, opened 2026-07-06
Feature/tiered storage lru policy

HoJacob:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

LRU initial implementation
### Checklist

- [ ] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #

### DIFF
--- a/component/tiered_storage/lru_policy.go
+++ b/component/tiered_storage/lru_policy.go
@@ -1,0 +1,69 @@
+
+package tiered_storage
+
+import (
+	"bytes"
+	"encoding/gob"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/Seagate/cloudfuse/common"
+	"github.com/Seagate/cloudfuse/common/log"
+)
+
+type lruNode struct{
+	prev *lruNode  
+	next *lruNode
+	name string 
+	
+}
+//upload 50 files and then check
+
+type lruQueue struct{
+	sync.Mutex
+
+	wg sync.WaitGroup
+	numWorkers int 
+
+	head *lruNode
+	tail *lruNode 
+
+	uploadChan chan string
+	doneChan chan struct{}
+
+	threshold float64
+	targetRatio float64
+
+
+}
+
+
+func(q *lruQueue) Touch(){}
+
+func(q *lruQueue) Add(){}
+
+func(q *lruQueue) Remove(){}
+
+//this will
+
+func worker(){}
+
+
+func (q *lruQueue) capacityChecker() {
+    defer q.wg.Done()
+    defer close(q.uploadChan) 
+
+    for {
+        select {
+        case <-time.After(//some time idk):
+            // eviction
+
+        case <-q.doneChan:
+            return
+        }
+    }
+}
--- a/component/tiered_storage/lru_policy.go
+++ b/component/tiered_storage/lru_policy.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/Seagate/cloudfuse/common"
@@ -24,39 +25,54 @@ type lruQueue struct {
 
 	nodeMap sync.Map
 
-	wg         sync.WaitGroup
-	numWorkers int
+	wg            sync.WaitGroup // tracks capacityChecker only
+	workerWg      sync.WaitGroup // tracks worker goroutines
+	numWorkers    int
+	activeWorkers int32
 
 	head *lruNode
 	tail *lruNode
 
-	uploadChan chan string
-	doneChan   chan struct{}
+	uploadChan   chan string
+	doneChan     chan struct{}
+	hallPassChan chan bool
 
-	cachePath    string
-	maxCacheSize float64
+	cachePath         string
+	maxCacheSize      float64
+	totalUploadedSize int64
 
 	threshold   float64
 	targetRatio float64
 
 	tickerUnit time.Duration
 
+	fileLocks *common.LockMap // uses object name (common.JoinUnixFilepath)
+
 	//Functions to wire later into tiered_storage package
 	//upload function from tiered storage WIRE THIS LATER in tiered storage because we are using a function from there
-	upload func(name string) error
+	uploadFn func(name string) error
 
+	//this function is just used for testing
 	FileHasOpenFileHandle func(name string) bool
 
 	//policy.isFileInUse = func(name string) bool {
 	//	return c.fileLocks.Get(name).Count() > 0
 	//}
 
-	//we also need the filLock map to use
+	//we also wire a cleanup function to delete from FileMap and Local
+	cleanupFn func(name string) error
+
+	// //delete locally
+	// 		localPath := filepath.Join(c.tmpPath, options.Name)
+	// 		c.mu.Lock()
+	// 		delete(c.fileMap, options.Name)
+	// 		c.mu.Unlock()
+	// 		os.Remove(localPath)
 
 }
 
 func (q *lruQueue) StartPolicy() error {
-	if q.upload == nil {
+	if q.uploadFn == nil {
 		return fmt.Errorf("lruQueue: upload function not set")
 	}
 	if q.numWorkers <= 0 {
@@ -66,23 +82,25 @@ func (q *lruQueue) StartPolicy() error {
 	q.head = nil
 	q.tail = nil
 	//channels
-	q.uploadChan = make(chan string, 1000)
 	q.doneChan = make(chan struct{})
+	q.hallPassChan = make(chan bool, 1)
+	q.hallPassChan <- true
+
 	//timer
 	//go routines
 	q.wg.Add(1)
 	go q.capacityChecker()
 
-	q.wg.Add(q.numWorkers)
-	for i := 0; i < q.numWorkers; i++ {
-		go q.worker()
-	}
 	return nil
 }
 
 func (q *lruQueue) StopPolicy() error {
 	close(q.doneChan)
+	// Wait for capacityChecker to exit — its deferred close(uploadChan) fires here,
+	// signalling workers that no more jobs are coming.
 	q.wg.Wait()
+	// Now wait for workers to drain whatever remains in uploadChan.
+	q.workerWg.Wait()
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


component/tiered_storage/lru_policy.go
```diff
@@ -0,0 +1,265 @@
...
+	maxCacheSize float64
+
+	threshold   float64
+	targetRatio float64
+
+	tickerUnit time.Duration
+
+	//Functions to wire later into tiered_storage package
+	//upload function from tiered storage WIRE THIS LATER in tiered storage because we are using a function from there
+	upload func(name string) error
```
foodprocessor: Since this is sort of a function pointer, should we name it `uploadFunction` or `uploadFn` or something?

component/tiered_storage/lru_policy.go
```diff
@@ -0,0 +1,265 @@
...
+			//find difference to evict down to 60%
+			difference := curSize - q.maxCacheSize*q.targetRatio
+			curEvictedSpace := 0
+			for curEvictedSpace < int(difference) {
+				nodeSize, evicted := q.eviction()
+				if !evicted {
+					break
+				}
+				curEvictedSpace += int(nodeSize)
+			}
```
foodprocessor, line 256: This looks good!
Later on, when we're looking at concurrency, let's make sure this code can't run more than once at a time (if the ticker is faster than eviction, we don't begin another eviction loop in parallel).

component/tiered_storage/lru_policy.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		log.Err("lruPolicy::capacityChecker : failed to stat file: %v", err)
+		q.mu.Unlock()
+		return 0, false
+	}
+	nodeSize := fileInfo.Size()
+
+	//remove node from queue and map
+	q.extractNode(nodeToEvict)
+	q.nodeMap.Delete(nodeToEvict.name)
```
foodprocessor: Is there a rule we want to follow, for what node membership in the map and the linked list *mean*? In other words, should we update our records and *then* execute the action (upload & delete), or visa versa? Which is better for error handling?

component/tiered_storage/lru_policy.go
```diff
@@ -0,0 +1,265 @@
...
+	if found {
+		// touch
+		q.extractNode(node)
+	} else {
+		// brand new node — update tail if list was empty
+		if q.tail == nil {
+			q.tail = node
+		}
+	}
+	q.setHead(node)
```
foodprocessor: This creates a new node before searching for an existing one. It still does what we expect in the end, but it smells off.

component/tiered_storage/lru_policy.go
```diff
@@ -0,0 +1,265 @@
...
+	//Maybe have a duplicate , that touches essentially
+
+	//lock earlier
+	q.mu.Lock()
+	defer q.mu.Unlock()
+
+	//create node
+	newNode := &lruNode{name: name}
+	val, found := q.nodeMap.LoadOrStore(name, newNode)
+	node := val.(*lruNode)
```
foodprocessor: Also, dereferencing / type asserting `val` before we know if `val` was found makes me nervous. I figure it's probably fine and just returns nil in practice, but from my experience with C, I see echoes of "nil pointer dereference" here.

foodprocessor (commented):
Very good work! Good design, solid choices.
The function pointers are a very clever way to maintain modularity. Nice touch!

component/tiered_storage/lru_policy.go
```diff
@@ -0,0 +1,265 @@
...
+
+	fileInfo, err := os.Stat(localPath)
+	if err != nil {
+		log.Err("lruPolicy::capacityChecker : failed to stat file: %v", err)
+		q.mu.Unlock()
+		return 0, false
+	}
+	nodeSize := fileInfo.Size()
+
+	//remove node from queue and map
```
jfantinhardesty: In the eviction code here, you delete from the LRU, but the file will still remain in the local cache? Do we want the file to stay in the cache or should we delete it when we evict?

component/tiered_storage/lru_policy_test.go
```diff
@@ -0,0 +1,334 @@
...
+	// file1 and file2 are the LRU tail so they should be evicted to reach targetRatio (60%)
+	suite.assert.Contains(snapshot, "file1")
+	suite.assert.Contains(snapshot, "file2")
+	// file3 and file4 are the most recently used, so they should NOT be evicted
+	suite.assert.NotContains(snapshot, "file3")
+	suite.assert.NotContains(snapshot, "file4")
+}
+
+// 6. Eviction, file with open handle, file with no open handle,
+func (suite *lruPolicyTestSuite) TestCapacityCheckerEvictionOpenHandle() {
```
jfantinhardesty, line 254: This test is a bit flaky. The sleep could probably be a bit longer. Here is how you can replicate it.

go test -run TestLRUPolicyTestSuite ./component/tiered_storage --tags=unittest,fuse3 -count=50

You should see by running the unit tests 50 times that this test and a few others may fail.

component/tiered_storage/lru_policy.go
```diff
@@ -0,0 +1,265 @@
...
+func (q *lruQueue) eviction() (int64, bool) {
+	q.mu.Lock()
+	nodeToEvict := q.tail
+	if nodeToEvict == nil {
+		q.mu.Unlock()
+		return 0, false
+	}
+
+	//find the first applicable node
+	for nodeToEvict != nil && q.FileHasOpenFileHandle(nodeToEvict.name) {
```
jfantinhardesty: I think there could be a race condition here. After this check, it is possible that a new request comes in that opens the file before it gets uploaded.

jfantinhardesty (changes requested):
Great work! This is a strong start to the LRU policy. Just have a few comments.

component/tiered_storage/tiered_storage.go
```diff
@@ -519,20 +622,34 @@ func (c *TieredStorage) ReleaseFile(options internal.ReleaseFileOptions) error {
...
-		c.mu.Lock()
-		node, exists := c.fileMap[options.Handle.Path]
-		c.mu.Unlock()
+		// c.mu.Lock()
+		// node, exists := c.fileMap[options.Handle.Path]
+		// c.mu.Unlock()
 
-		if !exists {
+		val, ok := c.fileMap.Load(options.Handle.Path)
+		node := val.(*FileNode)
```
jfantinhardesty: Here , the node := val.(*FileNode) should be after the if !ok check. This could cause a nil panic if the path is not found in the file map (unlikely).

component/tiered_storage/lru_policy.go
```diff
@@ -0,0 +1,358 @@
...
+
+		//1. Get handle count and file size
+		flock := q.fileLocks.Get(fileName)
+		flock.Lock()
+		handleCount := flock.Count()
+
+		localPath := filepath.Join(q.cachePath, fileName)
+		fileInfo, err := os.Stat(localPath)
+		if err != nil {
+			log.Err("lruPolicy::capacityChecker : failed to stat file: %v", err)
```
jfantinhardesty: If there is an error here, then the next line would have a nil panic (since the fileInfo would be null we couldn't get the size. This could maybe happen if the file was deleted before the worker uploaded it.

component/tiered_storage/tiered_storage.go
```diff
@@ -153,6 +169,20 @@ func (c *TieredStorage) Configure(_ bool) error {
...
+
+	//Wire in the LRU Policy
+	c.policy = &lruQueue{
+		cachePath:        c.tmpPath,
+		maxCacheSize:     c.maxCacheSize,
+		fileLocks:        c.fileLocks,
+		threshold:        0.8,
+		targetRatio:      0.6,
+		numWorkers:       8,
+		tickerUnit:       time.Millisecond,
```
jfantinhardesty, line 183: This ticker time is too small. This would check capacity every millisecond. This should be relatively large.

component/tiered_storage/tiered_storage.go
```diff
@@ -551,30 +669,22 @@ func (c *TieredStorage) ReleaseFile(options internal.ReleaseFileOptions) error {
...
+			// c.mu.Unlock()
+			c.fileMap.Delete(options.Handle.Path)
+
+			os.Remove(localPath)
 		} else {
-			//local only then just close the file, update LRU add to queue, we will get to this later
-			options.Handle.Cleanup()
+			// update LRU add to queue because cleaning up the file should be handled once the file is uploaded in LRU policy logic
+			if c.policy != nil {
+				c.policy.Enqueue(options.Handle.Path)
```
jfantinhardesty, line 676: There is a possible deadlock here. In release file, you hold flock, and then in Enqueue you take the mu lock. However, in the eviction code in lru_policy, eviction first locks mu and then calls flock. This would cause deadlock. The fix is to ensure when taking multiple locks, everyone takes them in the same order.

component/tiered_storage/tiered_storage.go
```diff
@@ -254,6 +288,56 @@ func (c *TieredStorage) CreateFile(
 }
 
 func (c *TieredStorage) DeleteFile(options internal.DeleteFileOptions) error {
```
jfantinhardesty, line 290: I think delete file should call policy.Dequeue to remove the file from the lru.

jfantinhardesty (changes requested):

component/tiered_storage/tiered_storage.go
```diff
@@ -551,30 +669,22 @@ func (c *TieredStorage) ReleaseFile(options internal.ReleaseFileOptions) error {
...
+			// c.mu.Lock()
+			// delete(c.fileMap, options.Handle.Path)
+			// c.mu.Unlock()
+			c.fileMap.Delete(options.Handle.Path)
+
+			os.Remove(localPath)
 		} else {
-			//local only then just close the file, update LRU add to queue, we will get to this later
-			options.Handle.Cleanup()
+			// update LRU add to queue because cleaning up the file should be handled once the file is uploaded in LRU policy logic
```
HoJacob, line 672: Ok here is a potential deadlock, James' comment got moved somewhere, but basically the order in which we take locks is inconsistent between locking the file and accessing the queue, for release/enqueue we hold flock then take mu lock in enqueue, but for eviction we do the opposite we lock mu then do fileLock, soooo we can either change ReleaseFile and DeleteFile, to unlock before enqueue and dequeue is called, or just change the eviction code

component/tiered_storage/lru_policy.go
```diff
@@ -0,0 +1,371 @@
...
+		}
+
+		fileSize := fileInfo.Size()
+
+		//2. Check if file eligible to upload
+		if handleCount == 0 {
+			err := q.uploadandCleanFn(fileName)
+			flock.Unlock()
+			//handle when file doesn't exist during upload we do not requeue otherwise we do enqueue
+			if err != nil {
```
HoJacob, line 351: Ok this existence error check might not be entirely useful because we lock the file prior to upload and unlock after, so deletion during upload shouldn't happen, because deleteFile waits on a file lock as well. Just a heads up in case you want to delete this code!

component/tiered_storage/tiered_storage.go
```diff
@@ -605,6 +708,21 @@ func (c *TieredStorage) uploadCachedFile(name string) error {
...
+	c.fileMap.Delete(name)
+	err = os.Remove(localPath)
+	if err != nil {
+		log.Err("TieredStorage::uploadandCleanFile : %s remove failed [%v]", name, err)
+		return err
+	}
+	return nil
+}
+
 func (c *TieredStorage) RenameFile(options internal.RenameFileOptions) error {
```
HoJacob, line 726: ok question, do we need to rename a cloud only file? if not we can just return here, its a design question

jfantinhardesty (approved):

PR #973, opened 2026-07-15
Improve S3 Directory Listing

jfantinhardesty:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

This PR introduces a new add-dir-markers CLI command to add directory markers to directories which currently don't have them. It also introduces a require-dir-markers S3 config option to improve directory listing performance by skipping the case to ListObjectsV2. This skips the known file extension work that had been done previously.

There is also an improvements to S3 RenameFile which now checks if the rename target is an existing directory and returns EISDIR.

### Checklist

- [x] Tested locally
- [ ] Added new dependencies
- [ ] Updated documentation
- [ ] Added tests

### Related Issues

- Related Issue #
- Closes #


cmd/add-dir-markers.go
foodprocessor, line 1: This looks good, but some comments would help a lot.

component/s3storage/client.go
```diff
@@ -617,81 +640,44 @@ func (cl *Client) getFileAttr(ctx context.Context, name string) (*internal.ObjAt
...
-			log.Err(
-				"Client::getDirectoryAttr : HeadObject(%s) failed. Here's why: %v",
-				dirName,
-				headErr,
-			)
-			return nil, headErr
-		}
+	if cl.Config.requireDirMarkers {
+		return cl.getDirMarkerAttr(ctx, dirName)
 	}
```
foodprocessor, line 647: If `requireDirMarkers` is false, but `enableDirMarker` is true, shouldn't we still call `getDirMarkerAttr`, just in case the marker is there? I ask because HeadObject is much cheaper than List, at least in terms of AWS API calls.

component/s3storage/client.go
```diff
@@ -617,81 +640,44 @@ func (cl *Client) getFileAttr(ctx context.Context, name string) (*internal.ObjAt
...
 	if len(objects) > 0 {
-		// create and return an objAttr for the directory
-		attr := internal.CreateObjAttrDir(dirName)
-		return attr, nil
+		return internal.CreateObjAttrDir(dirName), nil
+	}
+
+	if cl.Config.enableDirMarker {
+		return cl.getDirMarkerAttr(ctx, dirName)
 	}
```
foodprocessor, line 660: The previous comment is essentially my asking whether we should raise this call above the List call.
Let's be optimistic, and potentially save costly List calls which could be avoided when the marker is there.

component/s3storage/config.go
```diff
@@ -92,6 +93,13 @@ func ParseAndValidateConfig(s3 *S3Storage, opt Options, secrets ConfigSecrets) e
 	s3.stConfig.usePathStyle = opt.UsePathStyle
 	s3.stConfig.disableUsage = opt.DisableUsage
 	s3.stConfig.enableDirMarker = opt.EnableDirMarker
+	s3.stConfig.requireDirMarkers = opt.RequireDirMarkers
+	if opt.RequireDirMarkers && !opt.EnableDirMarker {
+		return fmt.Errorf(
+			"%w: require-dir-markers requires enable-dir-marker to be true",
+			errInvalidConfigField,
+		)
+	}
```
foodprocessor, line 102: Maybe we could make require override enable to true, and just print a warning. Since enable is false by default. If we want to be extra safe, we could override enable only when it's not explicitly set.

foodprocessor (approved):

PR #1016, opened 2026-08-22
Improve exception handling

foodprocessor:
Improve exception handling:

* Enforce display capacity <= total bucket capacity
* Don't ignore log output errors - they cause us to erase previous logs by rotating
* Back off on all cloud errors when uploading pending ops
* Retry failed uploads asynchronously, when permitted

Fix Windows permissions:

* Windows does not have a compatible permissions system, so don't let file permissions be set
* Set SDDL in a way that makes sense to explorer (add an owner and group)
* Include code to repair permissions on existing file cache files on Windows systems ("rogue read-only")
* Update copilot instructions

### DIFF
--- a/component/file_cache/async.go
+++ b/component/file_cache/async.go
@@ -38,6 +38,12 @@ import (
 	"github.com/netresearch/go-cron"
 )
 
+// bounds on the backoff between upload cycles that ended in failure
+const (
+	minRetryDelay = time.Second
+	maxRetryDelay = 15 * time.Second
+)
+
 type UploadWindow struct {
 	name        string `yaml:"name"`
 	cronExpr    string `yaml:"cron"`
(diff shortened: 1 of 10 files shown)


component/file_cache/async.go
```diff
@@ -295,8 +334,9 @@ func (fc *FileCache) updateObject(name string, flags pendingFlags) error {
 	// in case of inconsistency, local state takes precedence (except to prevent incorrect deletions)
 	if !flags.isDeletion && localErr != nil {
 		log.Err("FileCache::updateObject : %s stat failed. Here's why: %v", name, localErr)
+		// the op is resolved, not failed - dropping it must not end the cycle
```
jfantinhardesty: Not sure I follow this. Why should we drop the error here?

jfantinhardesty (approved):

component/file_cache/async.go
foodprocessor: What we're saying is that this file's failure should not affect whether we try to upload the next file in the loop.

foodprocessor: I see why this is confusing. I changed it to a boolean, which simply tells the calling loop whether to continue.

PR #1017, opened 2026-08-22
Fix windows permissions

foodprocessor:
* Change SDDL to make sense to Explorer
* Block all chmod calls from Windows
* Ignore CreateFile permissions bits on Windows (use PropFlagIsDefaultMode).
* Protect file cache against any read-only files from previous versions

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -1190,6 +1190,16 @@ func unlockAll(flocks []*common.LockMapItem) {
 	}
 }
 
+// cacheFileMode returns the mode to use for a local cache file. On platforms
+// without POSIX modes the caller's mode is meaningless and may leave the file
+// read-only, so the cache default is used instead.
+func (fc *FileCache) cacheFileMode(mode os.FileMode) os.FileMode {
+	if !posixFileModes {
+		return fc.defaultPermission
+	}
+	return mode
+}
+
 // CreateFile: Create the file in local cache.
 func (fc *FileCache) CreateFile(options internal.CreateFileOptions) (*handlemap.Handle, error) {
 	//defer exectime.StatTimeCurrentBlock("FileCache::CreateFile")()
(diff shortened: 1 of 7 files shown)


component/file_cache/file_cache_linux.go
```diff
@@ -38,6 +38,9 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+// posixFileModes reports whether the platform has meaningful POSIX file modes.
+const posixFileModes = true
```
jfantinhardesty: Not sure if this is more or less readable than doing this just this with runtime.GOOS.

component/file_cache/file_cache.go
```diff
@@ -1190,6 +1190,45 @@ func unlockAll(flocks []*common.LockMapItem) {
...
+// without POSIX modes the caller's mode is meaningless and may leave the file
+// read-only, so the cache default is used instead.
+func (fc *FileCache) cacheFileMode(mode os.FileMode) os.FileMode {
+	if !posixFileModes {
+		return fc.defaultPermission
+	}
+	return mode
+}
+
+// clearCacheFileReadOnly drops the read-only attribute that older Windows builds
```
jfantinhardesty, line 1203: Not in love with this fix, but I guess it is the only way to recover from files which were previously incorrectly set to read only.

jfantinhardesty (approved):

component/file_cache/file_cache_linux.go
foodprocessor: Yeah, I was unsure. I'll go ahead and drop the variable.

component/file_cache/file_cache.go
foodprocessor: I really don't like it either, but I think this is the way forward. The sooner this is deployed, the sooner we can justify ripping it out (hopefully, if ever).

PR #1018, opened 2026-08-24
Fix CI issues

foodprocessor:
Mostly AI changes aimed at improving CI reliability


jfantinhardesty (approved):

component/azstorage/block_blob_test.go
```diff
@@ -1768,7 +1798,9 @@ func (s *blockBlobTestSuite) TestTruncateEmptyFileToLargeSize() {
 
 	props, err := s.az.GetAttr(internal.GetAttrOptions{Name: name})
 	s.assert.NoError(err)
-	s.assert.NotNil(props)
+	if !s.assert.NotNil(props) {
```
jfantinhardesty: I don't think we should return if this fails, may hide future errors.

cmd/root.go
```diff
@@ -63,6 +63,7 @@ type Blob struct {
 }
 
 var disableVersionCheck bool
+var releaseAPIBaseURL = common.CloudfuseReleaseURL
```
jfantinhardesty, line 66: This introduced variable seems unneeded.

jfantinhardesty (approved):
Looks good, besides some minor things

component/azstorage/block_blob_test.go
foodprocessor: It'll still fail the test, right? And if GetAttr fails, we'll just get a nil pointer deref next.

cmd/root.go
foodprocessor: It's there to mask a constant, to allow for offline version checks in CI.

component/azstorage/block_blob_test.go
jfantinhardesty: Oh sure, yeah I guess you are right. I guess it just is a new pattern we haven't been doing. The assumption is the code passes all asserts. If it doesn't the test will fail and sometimes also get a nil pointer deref.

PR #1030, opened 2026-08-26
Fix config-base health monitor on Windows

foodprocessor:
Work harder to resolve cfusemon on the path, and disable monitoring correclty when it is not found.


jfantinhardesty (approved):
LGTM!

